PR #23, opened 2026-03-05
feat: add working stop button and send-while-streaming

### LINKED ISSUE #20: Chat: Stop button doesn't work while waiting for agent response
## Problem

When a user sends a chat message and waits for the agent to respond, they are blocked:

1. **Stop button doesn't work** — it exists in the UI (`ComposerPrimitive.Cancel`) but only manages client-side state. No abort signal is sent to OpenClaw, so the agent keeps running server-side.
2. **Can't send follow-up messages** — the composer is locked while `isRunning === true`.

This is frustrating because the stop button is exactly meant for this situation — and being unable to even queue a follow-up message feels broken.

## Root Cause

Pinchy's WebSocket runtime (`use-ws-runtime.ts`) processes one message at a time. When `isRunning` is `true`:
- The send button is hidden, replaced by the stop button
- The stop button triggers `ComposerPrimitive.Cancel` from assistant-ui, which resets client state but **never tells the server to abort**
- The server (`client-router.ts`) keeps consuming the OpenClaw stream until it finishes naturally

## What OpenClaw Actually Supports

Research into `openclaw-node` confirms full support for both abort and concurrent messages:

### `chatAbort(sessionKey, runId?)` 
Sends a `chat.abort` protocol message to the Gateway. Two-phase: (1) terminates the local JS generator immediately, (2) tells OpenClaw to stop the agent run. Already tested in openclaw-node's test suite.

### Concurrent messages
The Gateway protocol multiplexes via unique request IDs. Each `chat()` call gets its own stream, and responses are routed back by ID. Multiple in-flight requests on the same session are explicitly supported.

### Phase 1: Make the stop button work (bug fix)

1. **Add an `abort` message type** to the browser↔Pinchy WebSocket protocol
2. **Wire `ComposerPrimitive.Cancel`** to send `{ type: "abort" }` over the WebSocket
3. **In `client-router.ts`**, handle the `abort` message by calling `this.openclawClient.chatAbort(sessionKey)`
4. **Clean up client state** — set `isRunning = false`, allow new messages

This is the minimum fix. The stop button should actually stop the agent.

### Phase 2: Allow sending messages while agent is running (enhancement)

1. **Don't lock the composer** while `isRunning === true` — show both the send button and the stop button
2. **Queue or send concurrent messages** — OpenClaw supports this natively via multiplexed request IDs
3. **Handle overlapping streams** — each response maps to a request ID, so messages can be interleaved correctly
4. **UI consideration**: Show the user's queued message immediately in the chat, then stream the agent's responses in order

Phase 2 is a nice-to-have. Phase 1 (working stop button) is the priority.

## Key Files

- `packages/web/src/hooks/use-ws-runtime.ts` — client-side WebSocket runtime, `isRunning` state, `onNew` callback
- `packages/web/src/components/assistant-ui/thread.tsx` — stop button UI (`ComposerPrimitive.Cancel`)
- `packages/web/src/server/client-router.ts` — server-side message routing, OpenClaw stream consumption
- `packages/web/server.ts` — WebSocket server setup, message handler dispatch

clemenshelm:
## Summary

Closes #20

- **Server abort handler**: `client-router.ts` handles `abort` messages by calling `chatAbort()` on OpenClaw and responding with `aborted`
- **Client onCancel**: WebSocket runtime sends abort message, clears running/delay state and timers
- **Implicit abort**: Sending a new message while streaming automatically aborts the current stream first
- **UI**: Stop button moved from composer to the streaming assistant message (ChatGPT pattern); composer stays always active so users can type while the agent is responding

## Test Plan

- [x] 1208 unit tests passing
- [ ] Manual: Send a message, click Stop while streaming → stream stops
- [ ] Manual: Send a message while agent is streaming → old stream aborted, new message sent
- [ ] Manual: Reload after stopping → verify history consistency

### DIFF
--- a/packages/web/src/__tests__/hooks/use-ws-runtime.test.ts
+++ b/packages/web/src/__tests__/hooks/use-ws-runtime.test.ts
@@ -1182,6 +1182,10 @@ describe("useWsRuntime", () => {
       const abortMessage = JSON.parse(ws.send.mock.calls[2][0]);
       expect(abortMessage.type).toBe("abort");
       expect(abortMessage.agentId).toBe("agent-1");
+
+      // Should close the WebSocket to force reconnect
+      // (works around OpenClaw Gateway session lock bug after abort)
+      expect(ws.close).toHaveBeenCalled();
     });
 
     it("should set isRunning to false when onCancel is called", () => {
--- a/packages/web/src/hooks/use-ws-runtime.ts
+++ b/packages/web/src/hooks/use-ws-runtime.ts
@@ -78,6 +78,9 @@ export function useWsRuntime(agentId: string): {
   const reconnectAttemptRef = useRef(0);
   const reconnectTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
   const shouldRecoverFromHistoryRef = useRef(false);
+  // When true, the next reconnect keeps local messages instead of replacing
+  // them with server history. Used after abort to preserve partial responses.
+  const isAbortReconnectRef = useRef(false);
 
   // Reset state when switching agents — prevents stale messages from
   // one agent blocking history load for a different agent.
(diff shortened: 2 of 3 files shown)


clemenshelm:
## Closing — blocked on OpenClaw Gateway bug

This PR implemented a working stop button and abort flow, but testing revealed a fundamental bug in the OpenClaw Gateway: **the session lock is not released after `chat.abort`**, causing all subsequent messages on that session to return empty responses.

We attempted three server/client-side workarounds (client WS reconnect, stream draining, stream synchronization) — none resolved it because the lock is held inside the Gateway process itself.

### What was built (for reference if revisited later)
- Stop button in assistant message footer using `useThreadRuntime().cancelRun()`
- Server-side abort handler (`chatAbort` + `aborted` response)
- Client-side `messageId` filtering to ignore stale stream events
- Composer disabled while streaming (standard ChatGPT-like UX)
- pino structured logger replacing `console.error` in client-router
- Abort-reconnect workaround with partial response preservation

### Next step
Contribute a fix to OpenClaw Gateway for proper session lock cleanup after abort. Once that's merged, this feature can be re-implemented cleanly.

See [issue #20 comment](https://github.com/heypinchy/pinchy/issues/20#issuecomment-4010530520) for the full investigation.

PR #53, opened 2026-03-13
fix: use timingSafeEqual for gateway token validation

### LINKED ISSUE #44: security: use timingSafeEqual for gateway token validation
## Problem

`gateway-auth.ts` uses `===` for token comparison, which is vulnerable to timing attacks. An attacker could theoretically determine the gateway token character by character by measuring response times.

## Current code

```typescript
return token === gatewayToken;
```

## Fix

Use Node.js `crypto.timingSafeEqual()`:

```typescript
import { timingSafeEqual } from "crypto";

const tokenBuf = Buffer.from(token);
const gatewayBuf = Buffer.from(gatewayToken);
if (tokenBuf.length !== gatewayBuf.length) return false;
return timingSafeEqual(tokenBuf, gatewayBuf);
```

## Risk

Medium — the gateway is only reachable internally (Docker network), but defense in depth demands constant-time comparison for all secret comparisons.

elliotllliu:
## What does this PR do?

Replace `===` with `crypto.timingSafeEqual()` in `gateway-auth.ts` to prevent timing attacks on gateway token comparison.

Closes #44

## Type of change

- [x] 🐛 Bug fix

## Checklist

- [x] I've updated the documentation (if applicable)
- [x] All existing tests pass

## Details

The previous implementation used `===` for token comparison, which is vulnerable to timing attacks. An attacker could theoretically determine the gateway token character by character by measuring response times.

The fix uses Node.js `crypto.timingSafeEqual()` with a length pre-check (since `timingSafeEqual` requires equal-length buffers).

### DIFF
--- a/packages/web/src/lib/gateway-auth.ts
+++ b/packages/web/src/lib/gateway-auth.ts
@@ -1,4 +1,5 @@
 import { readFileSync } from "fs";
+import { timingSafeEqual } from "crypto";
 
 const CONFIG_PATH = process.env.OPENCLAW_CONFIG_PATH || "/openclaw-config/openclaw.json";
 
@@ -19,5 +20,8 @@ export function validateGatewayToken(headers: Headers): boolean {
   const gatewayToken = readGatewayToken();
   if (!gatewayToken) return false;
 
-  return token === gatewayToken;
+  const tokenBuf = Buffer.from(token);
+  const gatewayBuf = Buffer.from(gatewayToken);
+  if (tokenBuf.length !== gatewayBuf.length) return false;
+  return timingSafeEqual(tokenBuf, gatewayBuf);
 }


clemenshelm:
Thanks for the contribution!

However, this PR doesn't meet our quality bar:

- **Missing tests.** We practice TDD — all changes need tests, even small ones. A unit test for `validateGatewayToken()` with equal-length and different-length tokens would be straightforward.
- **Checklist inaccuracy.** "I've added tests for new functionality" is checked, but no tests were added.

We also ask that contributors open a [Discussion](https://github.com/heypinchy/pinchy/discussions) before submitting PRs, so we can align on approach first (see [CONTRIBUTING.md](https://github.com/heypinchy/pinchy/blob/main/CONTRIBUTING.md)).

We'll implement this fix ourselves when prioritized. Thanks for the interest in Pinchy!

PR #54, opened 2026-03-13
fix: make audit logging failures fail-closed with retry

### LINKED ISSUE #47: security: make audit logging failures non-silent
## Problem

The `pinchy-audit` plugin logs a warning when the audit endpoint fails, but the tool call still executes. This means agent actions can happen without being audited — a compliance gap.

## Current behavior

```typescript
// pinchy-audit/index.ts
logger?.warn?.("Failed to log tool use", error);
// Tool execution continues
```

## Options

1. **Fail-closed** — If audit logging fails, block the tool call. Strongest guarantee, but may cause availability issues if audit endpoint is down.
2. **Retry with fail-closed** — Retry audit log 2-3 times, then block if all fail.
3. **Configurable** — Env var `AUDIT_FAIL_MODE=closed|open` (default: `closed` in production).

## Risk

Medium — in regulated environments (finance, healthcare), unaudited actions are a compliance violation.

elliotllliu:
## What does this PR do?

Makes audit logging failures block tool execution by default, with configurable behavior via `AUDIT_FAIL_MODE` env var.

Closes #47

## Type of change

- [x] 🐛 Bug fix

## Checklist

- [x] I've updated the documentation (if applicable)
- [x] All existing tests pass

### Problem
When the audit endpoint fails, tool calls still execute — creating a compliance gap where agent actions happen without being audited.

### Solution
- **`AUDIT_FAIL_MODE=closed` (default)**: Block tool calls when audit logging fails. Fail-safe for compliance.
- **`AUDIT_FAIL_MODE=open`**: Warn and continue (previous behavior). For development or non-regulated environments.
- **Retry logic**: 2 retries with 500ms delay before failing — handles transient network issues without blocking unnecessarily.
- `before_tool_call` throws in closed mode to prevent unaudited execution.
- `after_tool_call` logs error but doesn't throw (tool already ran, blocking would be pointless).

### New tests
- Verify `open` mode doesn't throw on failure
- Verify `closed` mode throws on failure (default)
- Verify retry count (3 total attempts)

### DIFF
--- a/packages/plugins/pinchy-audit/index.test.ts
+++ b/packages/plugins/pinchy-audit/index.test.ts
@@ -1,4 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from "vitest";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 const mockOn = vi.fn();
 
@@ -30,6 +30,11 @@ describe("pinchy-audit plugin", () => {
   beforeEach(() => {
     vi.clearAllMocks();
     vi.stubGlobal("fetch", vi.fn().mockResolvedValue({ ok: true }));
+    delete process.env.AUDIT_FAIL_MODE;
+  });
+
+  afterEach(() => {
+    delete process.env.AUDIT_FAIL_MODE;
   });
 
   it("registers before_tool_call and after_tool_call hooks", async () => {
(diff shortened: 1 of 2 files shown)


clemenshelm:
Thanks for the contribution!

While this PR does include tests, there are several issues that prevent merging:

- **`sleep(500ms)` retry delay** blocks tool execution for up to 1.5s in the worst case — problematic for user experience.
- **`getAuditFailMode()` is called once at `register()` time**, not per-request. The env var `AUDIT_FAIL_MODE` can't be changed at runtime despite the docs suggesting otherwise.
- **Module caching in tests** — `vi.clearAllMocks()` + dynamic `import()` is fragile with Vitest's module cache. Tests may pass in isolation but fail or produce false positives when run together.

We also ask that contributors open a [Discussion](https://github.com/heypinchy/pinchy/discussions) before submitting PRs, so we can align on approach first (see [CONTRIBUTING.md](https://github.com/heypinchy/pinchy/blob/main/CONTRIBUTING.md)).

We'll implement this feature ourselves when prioritized. Thanks for the interest in Pinchy!

PR #144, opened 2026-04-16
docs: require SSH key at Hetzner setup + recovery path for legacy v0.3.0 installs

clemenshelm:
## Summary

Three problems surfaced while upgrading the production demo instance (`91.98.202.16`) from v0.3.0 to v0.4.0 strictly by following the published docs. This PR fixes all three so the next user — running v0.4.0-and-up — can upgrade in one command, and users stuck on a v0.3.0 install without SSH access get a clear recovery path.

### 1. `deploy-hetzner.mdx` step 7 — SSH key now required

The old step 7 reassured users "Skip SSH Keys (for now) — that's fine". The cloud-init intentionally doesn't set a root password either (that's correct — passwords on cloud images are a security liability). Net result: the server was left with no login path at all.

New step 7 makes the SSH key mandatory, embeds a `ssh-keygen` walkthrough for users without one, and a caution-aside that spells out exactly why skipping strands the server.

### 2. `upgrading.mdx` v0.3.0→v0.4.0 — recovery block + deployment-model note

Added "If you deployed v0.3.0 without an SSH key" steps: Hetzner Cloud Console → Reset root password → web console login → `ssh-import-id-gh YOUR_GITHUB_USERNAME` (shortest typeable command for the unreliable Hetzner web terminal) → normal SSH upgrade.

Also documents the quieter behaviour change: v0.4.0 switches from local `docker compose build` to pre-built GHCR images. `docker compose pull` handles it transparently, but users who see "Recreate" on all three services will now understand why.

### 3. `CONTRIBUTING.md` — one-time GHCR publish step

GHCR creates packages **private** on first push. The v0.4.0 release broke `docker compose pull` for everyone with `unauthorized` until we manually flipped visibility on `pinchy` and `pinchy-openclaw`. Noted in the release steps so the next maintainer adding a new image name knows. Not recurring — subsequent tag pushes inherit the visibility.

## Why this is docs-only

No code or cloud-init changes. The cloud-init is actually correct as-is (not setting a root password is the right thing to do); only the instructions around it were misleading.

## Test plan
- [x] Upgrade the demo instance end-to-end, documenting every deviation from the guide (this PR is the deviation list)
- [x] Recovery steps tested live: Hetzner password reset → web console → `ssh-import-id-gh clemenshelm` → SSH from Mac → standard upgrade path worked
- [ ] After merge: docs.heypinchy.com auto-deploys, spot-check that the MDX renders


fbglv:
I also discovered that Docker Compose – in my case Podman  Compose– doesn’t work correctly with the v0.4.0 release. When I tried the usual command: `podman compose up --build` only the `pinchy_db` container starts.

clemenshelm:
> I also discovered that Docker Compose – in my case Podman  Compose– doesn’t work correctly with the v0.4.0 release. When I tried the usual command: `podman compose up --build` only the `pinchy_db` container starts.

Thanks a lot for reporting this @fbglv . I could reproduce it and a fix is on the way (0.4.1).

fbglv:
> > I also discovered that Docker Compose – in my case Podman  Compose– doesn’t work correctly with the v0.4.0 release. When I tried the usual command: `podman compose up --build` only the `pinchy_db` container starts.
> 
> Thanks a lot for reporting this @fbglv . I could reproduce it and a fix is on the way (0.4.1).

Thank you @clemenshelm for the amazing project!

PR #161, opened 2026-04-19
fix: password reset shows correct form at /reset/[token]

clemenshelm:
## Summary

- Add new `/reset/[token]` page with password-only form ("Reset your password" heading, no name field) that submits to the existing `/api/invite/claim` endpoint
- Fix `user-detail-sheet.tsx` to generate reset links pointing to `/reset/` instead of `/invite/`
- Add `text-sm` to inline error paragraphs on both the reset and invite pages (project standard)

## Root Cause

When an admin triggered a password reset, the generated link pointed to `/invite/<token>` — which showed the account-creation form (name + password, "You've been invited to Pinchy") instead of a password reset form. No `/reset/[token]` page existed.

The backend (`/api/invite/claim`) already handled `type: "reset"` correctly — this was a pure frontend fix.

## Test Plan

- [ ] All 2,781 tests pass (`cd packages/web && npx vitest run`)
- [ ] Build succeeds with `/reset/[token]` as a dynamic route (`npx next build`)
- [ ] Admin triggers password reset → link shows `https://<host>/reset/<token>`
- [ ] Visiting `/reset/<token>` shows "Reset your password" form with Password + Confirm password fields (no Name field)
- [ ] Submitting valid matching passwords resets the password and shows "Password reset!" success screen
- [ ] "Continue to sign in" button redirects to `/login`
- [ ] Short password or mismatched passwords show inline validation errors

### DIFF
--- a/packages/web/e2e/08-password-reset.spec.ts
+++ b/packages/web/e2e/08-password-reset.spec.ts
@@ -1,0 +1,24 @@
+import { test, expect } from "@playwright/test";
+
+test.describe("Password reset page", () => {
+  test("/reset/[token] renders the reset form, not the invite form", async ({ page }) => {
+    await page.goto("/reset/some-token-value");
+
+    // Reset-specific copy
+    await expect(page.getByText("Reset your password")).toBeVisible();
+    await expect(page.getByText("Enter a new password for your account.")).toBeVisible();
+    await expect(page.getByRole("button", { name: /reset password/i })).toBeVisible();
+
+    // Must NOT render the invite-claim form (that was the bug)
+    await expect(page.getByText(/you've been invited to pinchy/i)).toHaveCount(0);
+    await expect(page.getByLabel(/^name$/i)).toHaveCount(0);
+  });
+
+  test("/invite/[token] still renders the invite form (regression guard)", async ({ page }) => {
+    await page.goto("/invite/some-token-value");
+
+    await expect(page.getByText("You've been invited to Pinchy")).toBeVisible();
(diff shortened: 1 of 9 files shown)


clemenshelm:
Thanks for this fix and the clear root-cause write-up. Closing it as **superseded** — the underlying bug is already resolved on `main`, but via a different architecture than the one here, so this PR can no longer merge cleanly and merging it would regress an intentional design decision.

What changed since this PR (last updated 2026-04-29):

- **The bug is fixed via a unified, type-aware page — not a separate `/reset/[token]` page.** `app/invite/[token]/page.tsx` now serves both flows from one route using an `InviteType = "invite" | "reset"` discriminator (#436). It fetches the token type (`data.type === "reset"`) and renders reset-specific copy ("Reset your Pinchy password", "Set a new password for your account.", "Reset password", no name field). An admin-triggered reset link therefore no longer shows the account-creation form. There is intentionally no `/reset/[token]` page on `main`.
- **Reset links intentionally stay on `/invite/`.** `lib/invite-url.ts#buildInviteUrl` returns `${origin}/invite/${token}` by design (see #352), and `user-detail-sheet.tsx` uses it for the reset link. This PR's switch to `/reset/` would revert that decision.
- `POST /api/invite/claim` also gained a hardened `type: "reset"` branch since (password hashing, session invalidation, an audit entry, and #436 name-overwrite protection) that this PR predates.

Net: the user-facing problem you reported is solved on `main`, and re-introducing the separate `/reset/` page would conflict with the consolidated flow. Closing — but thank you, the analysis here was spot-on and pointed at a real bug.

PR #210, opened 2026-04-30
fix(pinchy-odoo): write apiKey as plaintext, not SecretRef

### LINKED ISSUE #209: bug(pinchy-odoo): all tool calls fail with 'unhashable type: dict'
## Symptom

Reproduced on staging during v0.5.0 click-through with the CRM template (Piper, custom user agent). Every \`pinchy-odoo\` tool call fails with the same Python-side error:

\`\`\`
Error: unhashable type: 'dict'
\`\`\`

The agent (model: \`ollama-cloud/deepseek-v4-pro\`) makes valid calls — \`odoo_schema({model: "crm.lead"})\` and \`odoo_count({model: "crm.lead", filters: []})\` — and both return the error verbatim. The agent's reply (transcribed from the staging session JSONL):

> Leider bekomme ich bei allen Odoo-Abfragen mit Filtern den Fehler \`unhashable type: 'dict'\` — das deutet auf ein Backend-Problem im Odoo-Plugin hin. Nur \`odoo_schema\` ohne Parameter funktioniert.

## Tool-call evidence (from session JSONL)

Both calls are well-formed by the LLM. The empty-filter \`odoo_count\` rules out "LLM passed dict instead of tuple" as the root cause:

\`\`\`json
{ "name": "odoo_schema", "arguments": { "model": "crm.lead" } }
{ "name": "odoo_count",  "arguments": { "model": "crm.lead", "filters": [] } }
\`\`\`

Both results:
\`\`\`json
{ "role": "toolResult", "content": [{ "text": "Error: unhashable type: 'dict'" }] }
\`\`\`

## Likely root cause

The error originates from Python — Odoo's XML-RPC server rejects something in the request as unhashable. Since both tools fail identically (one without any filter at all), the bug lives in a shared layer:

- The request envelope generated by \`odoo-node\` (the library used in \`packages/plugins/pinchy-odoo/index.ts:1\`).
- Most likely: \`kwargs\` is being serialised as a positional argument instead of a keyword argument, so Odoo tries to use the dict as a hash key when matching against the method signature.

## Where to investigate

- \`packages/plugins/pinchy-odoo/index.ts:55\` (\`createClient\`) — what version of \`odoo-node\` are we on?
- \`packages/plugins/pinchy-odoo/index.ts:177\` (\`client.fields(model)\`) — first fail path; reproducible without filters.
- \`packages/plugins/pinchy-odoo/index.ts:291\` (\`client.searchCount(model, [])\`) — second fail path; same error confirms layer is shared.

## Reproduction (on a fresh staging deploy)

1. Configure an Odoo connection in admin Settings.
2. Create the CRM template (Piper) and grant it Odoo permissions.
3. Ask the agent any question that requires Odoo data, e.g. "Show me my most promising leads".
4. Inspect \`/root/.openclaw/agents/<id>/sessions/*.jsonl\` for \`toolResult\` entries — every \`pinchy-odoo\` call returns the same error.

## Severity

Blocks ALL Odoo-backed agent templates (Piper / Sales Analyst / Inventory Scout / etc.) on staging. The Odoo integration is one of v0.5.0's headline features — recommending we treat this as **0.5.0 release-blocker** unless a quick fix is found.

## Related

- #208 — separately, the same agent run on the broken \`llama3.3:70b\` resolver failed with HTTP 404 _before_ this Odoo bug surfaced. The model fix unblocks reaching this bug.

clemenshelm:
## Closes #209

User on staging: every \`pinchy-odoo\` tool call returned
\`Error: unhashable type: 'dict'\`. The Odoo template (Piper) was unusable; same for any other Odoo-backed agent.

## Root cause

\`openclaw-config.ts\` wrote \`pinchy-odoo.config.agents.<id>.connection.apiKey\` as a SecretRef object intending OpenClaw to resolve it at runtime. OpenClaw 2026.4.x only resolves SecretRefs in a hand-coded set of known config paths; it does not walk arbitrary plugin config trees. The unresolved dict reached \`pinchy-odoo\`, was forwarded to \`odoo-node\` as the password, and Odoo's Python server crashed when trying to use the dict as a hash key in its API-key lookup.

## Fix + four layers of guardrails

The runtime fix is in \`packages/web/src/lib/openclaw-config.ts\` — write \`apiKey\` as a plaintext string, same pragma already applied to \`pinchy-context.config.gatewayToken\` and Telegram bot tokens. Odoo API keys don't match any provider-key pattern in \`openclaw-plaintext-scanner.ts\` so the defense-in-depth check still passes.

**Layer 1 — Pinchy unit test (\`openclaw-config.test.ts\`).**
Asserts that \`apiKey\` lands in openclaw.json as a string. Regressing to the SecretRef shape fails this test before code is shipped.

**Layer 2 — Plugin runtime validation (\`pinchy-odoo/index.ts\`).**
\`createClient\` asserts \`connection.apiKey\` is a string and \`connection.uid\` is a number before instantiating \`OdooClient\`. If a future code path bypasses the type system (JSON without coercion, an unresolved SecretRef leaking through, etc.), users get a clear plugin-side error — \`apiKey must be a string, got object\` with a hint about #209 — instead of a confusing \`unhashable type: 'dict'\` from deep inside Odoo's Python.

**Layer 3 — Plugin unit tests for the validation (\`pinchy-odoo/__tests__/tools.test.ts\`).**
New \`connection shape validation (#209 guardrail)\` describe block: rejects SecretRef-shaped apiKey, rejects non-numeric uid, accepts well-formed config.

**Layer 4 — Plugin integration test against real mock-odoo (\`pinchy-odoo/__tests__/integration.test.ts\`).**
This file does NOT mock \`odoo-node\`. It boots an in-process instance of \`config/odoo-mock/server.js\` on an OS-picked port and exercises \`odoo_schema\`, \`odoo_count\`, \`odoo_read\`, plus a regression test that pushes the SecretRef-shaped dict through the full chain. The bug-shape test asserts the plugin fails fast and never produces the Python-server crash. Catches \`odoo-node\` serialisation issues that would slip past mocked unit tests.

\`config/odoo-mock/server.js\` exposes a \`start({ jsonRpcPort, controlPort })\` function so vitest can spin up isolated instances on ephemeral ports. CLI behaviour preserved via \`require.main\` check — the docker-compose service still auto-starts on 8069/9002.

## Tests

3344 tests pass locally (7 new). \`tsc --noEmit\` clean.

## Test plan

- [x] All tests pass locally + tsc clean
- [ ] CI green
- [ ] After merge + redeploy staging: Piper / Sales Analyst tool calls succeed (no more \`unhashable type: 'dict'\`)

## Related

- #208 — separate stale model ID fix (Piper got HTTP 404 *before* hitting this bug)
- #209 — root cause issue this PR closes

### DIFF
--- a/config/odoo-mock/server.js
+++ b/config/odoo-mock/server.js
@@ -764,14 +764,35 @@ const controlServer = http.createServer(async (req, res) => {
 });
 
 // ---------------------------------------------------------------------------
-// Start servers
+// Public API for in-process consumers (vitest integration tests).
+//
+// Auto-starts on the published Docker ports when run as a CLI (e.g. inside
+// the docker-compose service). Tests can instead import the module and
+// call `start()` to spin both servers up on ephemeral ports without
+// touching the docker-compose stack.
 // ---------------------------------------------------------------------------
 resetNextIds();
 
-jsonRpcServer.listen(8069, () => {
-  console.log("Mock Odoo JSON-RPC server listening on port 8069");
-});
+async function start({ jsonRpcPort = 8069, controlPort = 9002 } = {}) {
+  await new Promise((resolve) => jsonRpcServer.listen(jsonRpcPort, resolve));
+  await new Promise((resolve) => controlServer.listen(controlPort, resolve));
+  return {
+    jsonRpcPort: jsonRpcServer.address().port,
+    controlPort: controlServer.address().port,
+    async stop() {
+      await Promise.all([
+        new Promise((r) => jsonRpcServer.close(r)),
+        new Promise((r) => controlServer.close(r)),
+      ]);
+    },
+  };
+}
 
-controlServer.listen(9002, () => {
-  console.log("Mock Odoo Control API listening on port 9002");
-});
+if (require.main === module) {
+  start().then(({ jsonRpcPort, controlPort }) => {
+    console.log(`Mock Odoo JSON-RPC server listening on port ${jsonRpcPort}`);
+    console.log(`Mock Odoo Control API listening on port ${controlPort}`);
+  });
+}
+
+module.exports = { start };
--- a/packages/plugins/pinchy-odoo/__tests__/integration.test.ts
+++ b/packages/plugins/pinchy-odoo/__tests__/integration.test.ts
@@ -1,0 +1,161 @@
+/**
+ * Integration tests for pinchy-odoo against the real mock-odoo HTTP server.
+ *
+ * Layer 2 of the #209 guardrails: catches anything that breaks the actual
+ * plugin → odoo-node → Odoo JSON-RPC chain end-to-end. Layer 1 (unit tests
+ * in `tools.test.ts`) mock `odoo-node` away, so a regression in the way
+ * Pinchy's `apiKey` reaches Odoo (e.g. shape, encoding, header placement)
+ * would slip past them. This file uses the real `odoo-node` library
+ * against an in-process instance of `config/odoo-mock/server.js`.
+ *
+ * The original bug: Pinchy wrote the apiKey as a SecretRef object,
+ * `odoo-node` forwarded that dict to Odoo, and Odoo crashed with
+ * `unhashable type: 'dict'`. With this test in place, that exact
+ * regression — or any future variant where the wrong shape reaches
+ * Odoo over the wire — fails CI on a real RPC round-trip.
+ */
+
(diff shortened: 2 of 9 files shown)


clemenshelm:
Superseded by #211 — the right architectural fix is to fetch credentials from Pinchy on demand (Pattern B in CLAUDE.md), not write plaintext into openclaw.json. See discussion in #211.

Issue #193, opened 2026-04-29
Settings saves trigger full gateway restart cascade (15-30s downtime)


clemenshelm:
## Problem

Every settings save that touches `secrets.json` (provider API key, integration credentials, Telegram bot token, etc.) drops the active WebSocket connection for 15-45s. Reproduced on staging 2026-04-29:

```
15:48:44  User saves Ollama Cloud API key in Settings
15:48:44  Pinchy regenerateOpenClawConfig writes secrets.json (owned uid 999)
15:48:44  OpenClaw [SECRETS_RELOADER_DEGRADED] SecretProviderResolutionError:
          secrets.providers.pinchy.path must be owned by uid=0
15:48:59  start-openclaw.sh kills gateway (SIGTERM) — 15s
15:49:32  Pinchy reconnected (33s total)
```

Active chats are interrupted mid-conversation. Particularly bad during onboarding where the user is *expected* to configure providers.

## Why #190 doesn't cover this

Issue #190 (pre-create empty `secrets.json` stub) eliminates the **cold-start** mtime trigger. It does nothing for runtime writes — every settings save still produces a fresh, uid 999-owned `secrets.json`, OpenClaw's in-process reload still fails the ownership check, and `start-openclaw.sh`'s mtime watcher still kills the gateway.

## Root cause

OpenClaw 2026.4.12's strict ownership check: `secrets.providers.pinchy.path must be owned by the current user (uid=0)`. Pinchy can't write the file as root (uid 999, can't escalate). The `allowInsecurePath` flag is rejected by 2026.4.12's file-source schema (`additionalProperties: false`).

## Fix options

**a)** **Setuid chown helper.** A small SUID binary that Pinchy invokes after each `writeSecretsFile` to chown the file to root. Tightly scoped to one path; security-reviewable. Avoids the gateway restart entirely if OpenClaw's in-process reload then succeeds.

**b)** **Hand secrets writes to OpenClaw via WebSocket RPC.** Pinchy POSTs the bundle, OpenClaw (running as root) writes the file. Architectural change but the cleanest long-term — secrets never traverse Pinchy's filesystem.

**c)** **Wait for OpenClaw 2026.4.26+ with `allowInsecurePath` accepted on file-source.** Tracked separately; once available, drop the chown dance entirely.

(c) is preferred if/when the upstream change lands. (a) is the most Pinchy-isolated fix that ships today. (b) is the most invasive but eliminates the whole class of ownership races.

## Acceptance

[...]

PR #244, opened 2026-05-01
fix(openclaw-config): stable plugins.allow ordering to break #193 cascade

clemenshelm:
## Summary

- Roots out the `plugins.allow` cascade that's been failing `agent-create-no-restart.spec.ts` on main since PR #219
- Pinchy now produces a sorted, byte-stable `plugins.allow` array — no more "requires gateway restart" diffs against OpenClaw's auto-enabled writebacks

## Root cause

CI on main HEAD (`d091c6f`, run [25222463719](https://github.com/heypinchy/pinchy/actions/runs/25222463719)) shows the test failing with:

```
[reload] config change detected; evaluating reload (agents.list, plugins.allow)
[reload] config change requires gateway restart (plugins.allow)
[gateway] received SIGUSR1; restarting
```

This is a new cascade trigger that surfaced after PR #219 fixed `plugins.entries.telegram` non-idempotency. The previous code built `plugins.allow` as `[...new Set([...openClawPlugins, ...ourPlugins])]`, which puts non-pinchy plugins (telegram) FIRST and pinchy-* in entries insertion order. OpenClaw's auto-enable rewrites the array in alphabetical-ish order; Pinchy's next regenerate produces `[telegram, pinchy-docs, pinchy-context, pinchy-audit]` — same set, different order. OpenClaw's reload engine treats array order as significant → 15-30s "Agent runtime is not available" banner on every settings save.

## Fix

`.sort()` the final array. Stable, alphabetical, byte-identical across regenerate calls regardless of insertion order or what OpenClaw wrote back.

## Test plan

- [x] Unit test reproduces the exact CI fingerprint: existing `[pinchy-audit, pinchy-context, pinchy-docs, telegram]`, asserts written order matches
- [x] All 3390 existing tests still pass (`pnpm -C packages/web test`)
- [ ] Telegram E2E `agent-create-no-restart.spec.ts` turns green in CI
- [ ] Confirm on staging that no `requires gateway restart (plugins.allow)` lines appear after a settings save / agent create

Closes the dominant cause described in [#193 comment](https://github.com/heypinchy/pinchy/issues/193#issuecomment-4358266484) follow-up.

### DIFF
--- a/packages/web/src/__tests__/lib/openclaw-config.test.ts
+++ b/packages/web/src/__tests__/lib/openclaw-config.test.ts
@@ -2650,6 +2650,11 @@ describe("restart-state integration", () => {
     // diff'd against its own file sees `plugins.allow` as changed and triggers
     // a full gateway restart.
     expect(config.plugins.allow).toEqual(existingAllow);
+    // Stronger property: the output must be self-consistently sorted regardless
+    // of what OpenClaw wrote. Without this, the test above could pass simply
+    // because the test fixture happens to be alphabetical — masking the case
+    // where a future OpenClaw version uses a different deterministic order.
+    expect(config.plugins.allow).toEqual([...config.plugins.allow].sort());
   });
 
   it("regenerateOpenClawConfig is byte-idempotent against its own previous output (#193)", async () => {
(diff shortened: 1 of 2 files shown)


clemenshelm:
Superseded by [`348f7c64d`](https://github.com/heypinchy/pinchy/commit/348f7c64d) (merged before this PR's CI could finish).

Same root cause, different (and better) approach:
- This PR: `.sort()` the final `plugins.allow` array — alphabetical, depends on OpenClaw also writing alphabetically.
- Merged fix: position-preserving — walk `existing.plugins.allow`, keep wanted entries in their original positions, append new pinchy-* entries at the end. Byte-identical when the set is unchanged, so the early-return short-circuit kicks in and skips the write entirely.

Closing without merge. The merged commit also carries an equivalent regression test.

PR #267, opened 2026-05-03
chore(deps): bump openclaw 2026.4.12 → 2026.4.14

clemenshelm:
## Summary

Bump `openclaw` from `2026.4.12` to `2026.4.14`.

## Why this version (and not later)

Going further hits separate breaking changes that don't belong in this PR:

| Version | Issue |
|---------|-------|
| 4.15 | Provider/auth resolution moves to per-agent `auth-profiles.json`; `regenerateOpenClawConfig()` doesn't yet write those, so the integration agent-chat times out (`Integration test response.` never streams back through OpenClaw → fake-Ollama). |
| 4.29 | `device.pair.approve` is gated on `operator.admin` scope (CLI auto-approve loop in `config/start-openclaw.sh` no longer authorised). Plus `resolvePairingLocality` classifies container peers as `remote` and silently rejects operator pairings — Pinchy ends up in a `closed before connect` loop. Needs an `openclaw-node` patch to handle `pairing-required` close reason. |

The bisection sequence is recorded in commit history if useful; see prior pushes on this branch.

## What this picks up

The upstream fix for `openclaw#47458` (polling stall loop on multi-bot channel restart) landed before 4.20, so it's not in 4.14 — but the 4.13→4.14 delta still includes a few smaller stability fixes the gateway log was nagging us about. We'll get the polling-stall fix in a follow-up once the 4.15+ provider/auth and 4.29+ pairing-locality regressions are solved.

## Test plan

- [x] CI fully green (Lint/Test/Build, Vitest Integration, Integration OpenClaw+Fake Ollama, Docker smoke, E2E, End-user install/upgrade, Telegram E2E, Odoo E2E, Ollama integration, License Keypair, Repo hygiene, Security audit)
- [ ] Manual smoke test by @clemenshelm

### DIFF
--- a/Dockerfile.openclaw
+++ b/Dockerfile.openclaw
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y \
       build-essential \
     && rm -rf /var/lib/apt/lists/*
 
-RUN npm install -g openclaw@2026.4.14
+RUN npm install -g openclaw@2026.4.27
 
 # Install pinchy-files plugin dependencies (native modules must be built in the container)
 COPY packages/plugins/pinchy-files/package.json /tmp/pinchy-files/package.json
--- a/Dockerfile.openclaw
+++ b/Dockerfile.openclaw
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y \
       build-essential \
     && rm -rf /var/lib/apt/lists/*
 
-RUN npm install -g openclaw@2026.4.12
+RUN npm install -g openclaw@2026.4.14
 
 # Install pinchy-files plugin dependencies (native modules must be built in the container)
 COPY packages/plugins/pinchy-files/package.json /tmp/pinchy-files/package.json
--- a/Dockerfile.openclaw
+++ b/Dockerfile.openclaw
@@ -15,7 +15,7 @@ RUN apt-get update && apt-get install -y \
       build-essential \
     && rm -rf /var/lib/apt/lists/*
 
-RUN npm install -g openclaw@2026.4.14
+RUN npm install -g openclaw@2026.4.15
 
 # Install pinchy-files plugin dependencies (native modules must be built in the container)
 COPY packages/plugins/pinchy-files/package.json /tmp/pinchy-files/package.json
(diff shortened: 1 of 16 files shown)


clemenshelm:
## Status update — closing without merge

I went deep on this and **the upgrade path is not safe to land in this PR**. Summary of what I found:

### What works
- `openclaw-node 0.7.0` connects successfully to `openclaw 2026.5.2` in **direct loopback** (my macOS host → local gateway). All connection tests pass, sessions list, health, etc.

### What breaks in CI (Docker, container-to-container)
Pinchy → OpenClaw connect closes with `code=1008 reason=n/a` on every attempt, even after I tried:

1. Bumping scopes from `["operator.admin"]` → `["operator.read", "operator.write", "operator.admin"]` → `["operator.read", "operator.write"]`
2. Pinning to `2026.4.27` (last release before the 4.29 scope tightening)
3. Removing the `auto_approve_devices` Bash CLI loop (its `device.pair.approve` calls fail with `missing scope: operator.admin` since 4.29)
4. Putting it back

### Root cause (likely)
Reading the OpenClaw server source from the npm tarball, `resolvePairingLocality` classifies any non-loopback peer as `"remote"`, and `shouldAllowSilentLocalPairing` rejects `remote` outright. That makes every fresh Pinchy pair request **non-silent**, which closes the WS with 1008 (`pairing-required-…`) until an external admin approves. The old Bash loop that called `openclaw devices approve --latest` was the external admin — but on 4.29+ that RPC requires `operator.admin` scope which the gateway-shared token doesn't grant the CLI's connection.

The `gateway.nodes.pairing.autoApproveCidrs` knob exists but is explicitly node-role-only, not operator-role. There is no equivalent operator-role auto-approve config I can find in the schema.

### Likely real fix (not done here)
1. **Either** teach `openclaw-node` to handle the `pairing-required` close gracefully — read the pending request id from the close reason and surface it to Pinchy so we can drive the approval through a different code path.
2. **Or** wait for upstream OpenClaw to add an operator-equivalent of `gateway.nodes.pairing.autoApproveCidrs` so we can allowlist the Pinchy → OpenClaw container subnet.
3. **Or** request the bootstrap-pair-token / shared-secret-loopback-local path explicitly from openclaw-node so the gateway treats Pinchy as `cli_container_local` (which would auto-approve).

I lean (3) — needs an openclaw-node change but it's the most surgical.

Closing this and tracking it as an issue.

PR #415, opened 2026-05-21
[VERIFY-ONLY] test(e2e): cursor-jump regression guard (expected to fail)

clemenshelm:
**Do not merge.** This is a validation PR: it adds ONLY the new e2e regression guard for the v0.5.4 cursor-jump bug, on top of `main` (without the fix). If the test correctly catches the bug, CI's **E2E Tests** job should fail with:

```
e2e/18-composer-cursor.spec.ts > Composer cursor preservation > inserting text mid-string...
Expected: "helloXYZ world"
Received: "helloX worldYZ"
```

The actual fix + the same test live in PR #414. Once this PR's CI demonstrates the failure, close without merging.

### DIFF
--- a/packages/web/e2e/18-composer-cursor.spec.ts
+++ b/packages/web/e2e/18-composer-cursor.spec.ts
@@ -50,14 +50,19 @@ test.describe("Composer cursor preservation", () => {
     await page.keyboard.type("hello world");
     await expect(input).toHaveValue("hello world");
 
-    // Position the caret between "hello" and " world" (index 5). Setting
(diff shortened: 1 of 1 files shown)


clemenshelm:
Closing without merge. The e2e regression guard this branch verified turned out to not catch the actual cursor-jump bug — it reliably failed on both no-fix and fix branches but with different symptom signatures than the user's reported behavior. The test exercises a different composer interaction path than real-user typing. PR #414 ships the wrapper revert without the test; user will verify the fix manually on staging.

PR #463, opened 2026-06-02
fix(setup-wizard-e2e): settle gateway before first chat to kill flaky timeout

clemenshelm:
## The flake

The Setup Wizard E2E job intermittently fails with a 90 s timeout on `helpers.ts:229` (`/sure, happy to help/` never appears), and **the failing provider varies per run** — a tell-tale moving flake:

| Run | Anthropic | Google | Ollama Cloud | Ollama Local | OpenAI |
|---|---|---|---|---|---|
| main `c6b0d948` | ✓ | ✘ | ✓ | ✓ | ✘ |
| PR #459 rebased | ✓ | ✓ | ✘ | ✓ | ✓ |

Same helper, same assertion, different random provider — so it is not provider-specific logic, it is timing.

## Root cause

`runProviderSmokeTest` Phase 4 sends the first chat message **into an in-flight gateway restart**:

1. Phase 3 "Continue" saves the provider key → `regenerateOpenClawConfig` writes `secrets.json` for the first time.
2. `config/start-openclaw.sh` pkills + respawns the gateway on that first-secrets bootstrap marker (~40 s respawn).
3. Phase 4 immediately sends "Hello…" and waits a single flat 90 s, trusting dispatch-retry to absorb the restart.

`resetStack` already settles OpenClaw **before** the wizard, but the bootstrap pkill happens **during** the wizard (Phase 3), after that settle — so nothing fences it off. Under CI load (5 specs sequential, `workers:1`, `retries:0`) the respawn + round-trip occasionally overruns 90 s for whichever provider draws the slow restart.

## Fix — fence the race, don't widen the window

Wait for the gateway to reconnect (the same `/api/health/openclaw` settle poll `resetStack` uses) **before** sending the first message, so it goes to a healthy gateway and the reply is a single fast round-trip.

This **redistributes** the former 90 s Phase 4 budget rather than enlarging it:
- ≤60 s draining the bootstrap restart (condition-based, cheap health poll), then
- 30 s reply window against an already-healthy gateway.

Total Phase 4 ceiling stays ≤90 s, well under the existing 120 s per-test timeout — **no timeout bump, no retries, no weakened assertion**. The inline settle loop is extracted as `waitForOpenClawSettled`, now shared by `resetStack` and `runProviderSmokeTest`.

## Verification note

This is an E2E-infra reliability fix; it can't be reproduced deterministically locally (needs the full Docker stack + mock providers + CI-level load). The change is grounded in the documented pkill/respawn mechanism and reuses the already-proven settle logic. CI on this branch exercises it; the moving-flake signature should disappear.

Unblocks the Setup Wizard E2E check that was intermittently failing on #459 and on main.

### DIFF
--- a/packages/web/e2e/setup-wizard/helpers.ts
+++ b/packages/web/e2e/setup-wizard/helpers.ts
@@ -115,9 +115,21 @@ export async function resetStack(): Promise<void> {
   // (a transient reload/restart resets the streak). 60 s budget covers a
   // worst-case cold-start reload cycle while leaving headroom under the 120 s
   // beforeAll timeout (this runs after the ~30 s container restart above).
-  const settleDeadline = Date.now() + 60000;
+  await waitForOpenClawSettled(60000, "after reset");
+}
+
+/**
+ * Poll Pinchy's `/api/health/openclaw` until OpenClaw reports connected & ok
+ * continuously for 5 s (a transient reload/restart resets the streak), or throw
+ * after `budgetMs`. Fences a chat round-trip off from an in-flight gateway
+ * restart so the test never races a respawn cycle. Shared by resetStack (drains
+ * the cold-start cascade before the wizard) and runProviderSmokeTest (drains the
+ * first-secrets.json bootstrap pkill before the first message).
+ */
+export async function waitForOpenClawSettled(budgetMs: number, context: string): Promise<void> {
+  const deadline = Date.now() + budgetMs;
   let connectedSince: number | null = null;
-  while (Date.now() < settleDeadline) {
+  while (Date.now() < deadline) {
     try {
       const out = execSync(`curl -fsS http://localhost:7777/api/health/openclaw`, {
         stdio: "pipe",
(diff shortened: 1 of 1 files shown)


clemenshelm:
Closing as superseded by the setup-wizard E2E work now on `main` (the #464 line).

`main` already contains a strictly better version of this PR's idea:

- **Settle-before-first-chat** — `waitForOpenClawSettledViaPage(page, { stableForMs: 12000, deadlineMs: 90000 })` in `runProviderSmokeTest`, the same "don't dispatch into the in-flight gateway restart" approach as here, but page-context based with a 12 s stability streak (vs the 5 s here), plus a `resetStack` rewrite that stops OpenClaw before the Pinchy restart to avoid the size-drop SIGUSR1 cascade entirely.
- **The real root cause** — #464 fixes the dropped `config.apply` on OC rate-limit (`unknown agent id`), which is what this PR's failing Web Search E2E run actually hit. A gateway-connection settle (this PR) doesn't detect the agent-not-in-runtime state; the config-propagation fix does.
- **Correct reply budget** — `main` uses a 160 s assertion to outlast the server-side `chatWithDispatchRaceRetry` (150 s), because the first cold provider spec's agent apply can lag ~108 s. This PR reduced the reply window to 30 s, which would have *regressed* the test under exactly that cold-start lag.

No additional benefit remains here, and the merge conflict is with `main`'s superior implementation. Closing.

PR #470, opened 2026-06-08
fix(chat): streaming-resume duplicate message-id crash

clemenshelm:
Fixes the streaming-resume duplicate-message-id crash found in the staging click-through (Phase 3).

## Bug
Reloading the page **during** a streaming reply hard-crashes the whole chat view: assistant-ui's MessageRepository throws `a message with the same id already exists in the parent tree` and the app error boundary shows `Something went wrong`. A second reload (stream finished) shows the full reply.

## Root cause
On reload-mid-stream the history frame carries an `activeRun` signal and the reconcile relabels the last history assistant message to `activeRun.messageId`. When the in-flight assistant isn't in history yet (OpenClaw hasn't persisted it), the relabel lands on a prior assistant and the **chunk handler — which only checked the trailing message — appended a SECOND message with the run id**, the duplicate that crashes assistant-ui.

## Layered fix
- **Layer 1 — crash guard:** `dedupeById()` collapses duplicate ids right before `useExternalStoreRuntime`, so a transient duplicate can never crash the view. Pure, unit-tested (5 cases).
- **Layer 2 — root cause:** `mergeOrAppendChunk()` merges a chunk into the assistant message carrying its id **anywhere** in the list (not just the tail), so the duplicate is never created. Pure, unit-tested (4 cases incl. the non-last resume case). 219 hook tests green, no streaming regression.
- **Layer 3 — E2E:** Playwright reload-mid-stream regression — reload the same tab while streaming, assert no error boundary and that the reply resumes in one bubble.

Deterministic red/green proof is at the unit layer (Layers 1+2); the E2E is the end-to-end behavior guard (reliably green with the fix). Build/typecheck green.

### DIFF
--- a/packages/web/e2e/integration/17-reload-mid-stream-resume.spec.ts
+++ b/packages/web/e2e/integration/17-reload-mid-stream-resume.spec.ts
@@ -1,0 +1,75 @@
+// packages/web/e2e/integration/17-reload-mid-stream-resume.spec.ts
+//
+// Regression guard for the streaming-resume duplicate-message-id crash.
+//
+// Reloading the SAME tab while a reply is still streaming exercises the
+// `activeRun` resume path: the reconnecting client fetches history + an
+// activeRun signal and re-attaches to the in-flight run. A prior bug let the
+// in-flight assistant message be materialised twice with the same id, which
+// crashes assistant-ui's MessageRepository and replaces the chat with the
+// "Something went wrong" error boundary. (Fixed by dedupeById + mergeOrAppendChunk.)
+//
+// With the fix this is reliably green: whether the reload lands on the resume
+// path (run still in-flight) or the history path (run finished during reload),
+// the view must NOT crash and the reply must complete in a single bubble.
+import { test, expect } from "@playwright/test";
+import {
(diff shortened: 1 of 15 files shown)


clemenshelm:
⚠️ **Layers 1+2 do NOT fully fix the crash — back to draft.**

The Layer-3 E2E (reload-mid-stream) correctly fails: after reload the `Something went wrong` error boundary still appears (persistently, 14/14 polls). So the dedupe guard + merge-by-id are necessary defense-in-depth / correctness improvements, but they are **not** the complete fix.

This is the E2E earning its keep — it caught a false fix that the green unit tests alone would have shipped.

**Likely real root cause (unconfirmed):** assistant-ui re-links the stable `activeRun.messageId` across history reconciles that reassign fresh `uuid()`s to every other message — an across-render re-link, not a within-array duplicate (which is all dedupeById catches). Needs local browser reproduction (integration stack + console) to confirm the exact MessageRepository.performOp('link') throw condition and fix it properly (probably: keep message ids stable across reconciles).

Next: reproduce locally, confirm root cause, then fix. Not merging until the E2E is green.

Issue #370, opened 2026-05-18
Docker image disk-hygiene: prune stale layers on upgrade


clemenshelm:
## Problem

`docker compose pull` keeps every previous layer of `pinchy:*` and `pinchy-openclaw:*` around until something explicitly prunes them. A long-running deployment that takes a handful of upgrades — especially one that tracks `:next` pre-releases — quickly accumulates dangling images and runs out of disk.

We hit this on `staging.heypinchy.com` today: 75 GB root volume, 100 % full, with 29 Docker images totalling ~73 GB (most of them `<none>:<none>` layers from prior pulls). The `pinchy` container crashed mid-startup with `ENOSPC: no space left on device` while trying to run database migrations (`mkdir '/tmp/tsx-999'`), restart-looped, and the deploy aborted with `dependency failed to start: container pinchy-pinchy-1 is unhealthy`.

Manual workaround: `docker image prune -a -f` (recovered ~60 GB instantly).

This isn't urgent for typical customers — they upgrade every few weeks at most, between tagged releases — but it's a latent landmine that will eventually bite anyone who runs Pinchy on a small VM for a long time. It already bites our own staging because we deploy many pre-releases per day.

## Proposal

Wire an opt-out disk-hygiene step into the standard upgrade flow. Candidate approaches, in order of how invasive they are:

1. **Document only** — note in `docs/src/content/docs/guides/upgrading.mdx` that customers running on small disks should periodically run `docker image prune -a -f` between upgrades. Low effort, easy to miss.
2. **Add prune to the upgrade snippet** — change the canonical `cd /opt/pinchy && docker compose pull && docker compose up -d` line to include `docker image prune -f` before `up -d`. Removes only dangling layers; safer than `-a`.
3. **Periodic cron** — a systemd timer or cron entry installed alongside Pinchy that runs `docker image prune -f` weekly. Most reliable; needs install/uninstall plumbing.
4. **Compose-level retention policy** — keep the N most recent tags of `pinchy*` images and drop older ones. Most precise; most code.

Recommend (2) plus a one-paragraph note in the upgrading guide explaining when to escalate to `-a -f` (after several version bumps or before a big release window).

## Acceptance criteria

- Customers following the documented upgrade flow on a 20 GB disk over six version bumps stay under 80 % disk use.
- The hygiene step is idempotent and safe to run on a running deployment.
- Upgrading guide covers when manual `prune -a` is appropriate vs. the default `prune`.

PR #487, opened 2026-06-11
feat(security): fail closed on default DB password + secret provenance in /api/health & Settings

### LINKED ISSUE #156: Hardening: fail-closed on default DB_PASSWORD + expose secret source in /api/health & Settings
Follow-up to #155. Two small hardening steps that are independently useful and share the same surface area (startup-time config validation + operator-facing observability).

## 1. Fail-closed on default DB_PASSWORD in production

`server.ts:64-71` currently only *warns* when `DATABASE_URL` still contains the default `pinchy_dev` password:

\`\`\`
WARNING: Using default DB_PASSWORD. Set a secure password via .env for production.
\`\`\`

Postgres is only reachable inside the Docker network, so the blast radius is lower than the other secrets — but a CISO reviewing logs will still flag it. Convert the warn to a hard exit in production (mirror the pattern used for other hard-fail config). In dev (`NODE_ENV !== \"production\"`), keep the warning so the dev-compose smoke tests don't break.

## 2. Expose \"secrets source\" in `/api/health` and the Settings UI

`docker inspect` shows env vars as empty strings when only the file fallback is active (auto-generated secret in `/app/secrets/.encryption_key` / `.better_auth_secret`). During the demo HTTPS setup I misread this as \"no secret\" and rotated secrets that didn't need rotating, causing data loss. The auto-gen fallback is correct and documented in code, but invisible from outside the container.

Add a non-secret-revealing \"source\" indicator:

- `/api/health` JSON gains a `secrets` object: `{ encryption_key: \"envvar\"|\"file\"|\"unset\", auth_secret: \"envvar\"|\"file\"|\"unset\", db_password: \"custom\"|\"default\" }`. No values, just provenance.
- Settings → Security (or a new \"Diagnostics\" panel) displays the same, with a hint text like *\"Encryption key loaded from: persisted file (Docker volume). Environment variable overrides this.\"*

This would have saved me from my own mistake today and will save operators from the same trap.

## Acceptance criteria

- [ ] Starting Pinchy in `NODE_ENV=production` with `DATABASE_URL=...pinchy_dev...` exits non-zero with a clear error message. Starting in development still only warns.
- [ ] `GET /api/health` returns a `secrets` object with the three provenance fields (no secret values ever).
- [ ] Tests cover all three states for each secret (envvar / file / unset).
- [ ] Settings UI has a visible \"secret source\" indicator — doesn't need to be fancy, just truthful.

Refs: #155 (secure-by-default deployment)

clemenshelm:
Closes #156.

## What

Two hardening steps from the issue, sharing the same surface (startup config validation + operator-facing observability):

### 1. Fail-closed on default DB_PASSWORD in production

`server.ts` previously only logged `WARNING: Using default DB_PASSWORD` and started anyway. Now:

- **production + default password → `console.error` + `process.exit(1)`** with a message that says exactly what to do.
- **dev + default password → warning** (unchanged behavior for dev-compose smoke tests).

The decision table is a pure function (`evaluateDbPasswordPolicy` in `src/lib/secret-source.ts`) so the policy is unit-tested and the server wiring stays a one-liner.

### 2. Secret provenance in /api/health and Settings → Security

`GET /api/health` now returns:

```json
{ "status": "ok", "secrets": { "encryption_key": "envvar|file|unset", "auth_secret": "envvar|unset", "audit_hmac_secret": "envvar|file|unset", "db_password": "custom|default" } }
```

Provenance only — no values. Settings → Security shows the same as a new **Secrets** card (own component, so the existing `SettingsSecurity` fetch-sequence tests stay untouched), with an amber flag on the default DB password. `audit_hmac_secret` is included beyond the issue's list since it's the third file-backed secret with the same operator-confusion potential.

To keep provenance and actual secret resolution from drifting, `getOrCreateSecret()` was refactored to dispatch on the new `getSecretSource()` — one priority order, two consumers.

## Deployment impact (please read)

This is a **breaking change for production deployments that never set `DB_PASSWORD`** — they will refuse to start after upgrading. `docs/guides/upgrading.mdx` has a new section with the recovery recipe (ALTER USER first — `POSTGRES_PASSWORD` only applies on first init — then `.env`, then upgrade). **Check demo.heypinchy.com and pinchy.heypinchy.com before tagging the next release.**

Security trade-off note: `/api/health` is unauthenticated; the provenance object reveals config posture (e.g. `db_password: default`) but no secret material. The issue explicitly wants the shell-curl use case; the default password is public in the repo anyway, and the DB is not network-exposed. Flagging it in case you'd rather gate the field on an admin session.

Also pins the v0.5.6→v0.5.7 `upgrading.mdx` heading/placeholders to v0.5.7 — they were left unpinned by the release, and the docs build resolves `%%PINCHY_VERSION%%` with the *current* version, which would have mislabeled that section after the next release.

## Tests

TDD throughout (all watched red first):
- `secret-source.test.ts` — source resolution incl. invalid-env fallthrough, auth-secret whitespace handling, policy decision table.
- `health.test.ts` — provenance shape + an assert that no secret values appear anywhere in the response body.
- `secrets-provenance-card.test.tsx` — provenance rendering, default-password flag, calm copy for not-yet-created secrets, renders nothing on older servers/fetch failure.
- Full suite: 5720 passed. Docs build clean (incl. the #370 prune guard on the new upgrade section).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,10 +294,11 @@ jobs:
       packages: read
     # Dev-mode CI test: images are built locally and tagged :latest. Satisfy
     # the compose fail-loud check on ${PINCHY_VERSION:?} without depending on
-    # a .env file.
+    # a .env file. Deliberately NO DB_PASSWORD: this stack exercises the
+    # entrypoint's password auto-migration (#156) like a real install that
+    # never set one — asserted below via /api/health secrets provenance.
     env:
       PINCHY_VERSION: latest
-      DB_PASSWORD: pinchy-ci-smoke-pass
 
     steps:
       - uses: actions/checkout@v4
@@ -403,6 +404,24 @@ jobs:
           echo "::error::Infrastructure not ready: database=$DB, openclaw=$OC"
           docker compose logs pinchy openclaw
           exit 1
+
+      - name: Verify DB password auto-migration (#156)
+        run: |
+          # The stack booted without DB_PASSWORD, so the entrypoint must have
+          # migrated the database off the public default and persisted the
+          # generated credential in the secrets volume.
+          PROV=$(curl -sf http://localhost:7777/api/health | jq -r '.secrets.db_password')
+          if [ "$PROV" != "generated" ]; then
+            echo "::error::expected secrets.db_password=generated after auto-migration, got: $PROV"
+            docker compose logs pinchy | grep -i -E "(db-password|password)" || true
+            exit 1
+          fi
+          docker compose logs pinchy | grep -q "Database is using a managed (non-default) password" || {
+            echo "::error::entrypoint migration log line missing"
+            docker compose logs pinchy | head -40
+            exit 1
+          }
+          echo "DB password auto-migration verified (provenance: generated)"
 
       - name: Run setup wizard (production)
         run: |
@@ -536,6 +555,16 @@ jobs:
           docker compose ps
           docker compose logs
           exit 1
+
+      - name: Verify generated DB password survives recreate (#156)
+        run: |
+          PROV=$(curl -sf http://localhost:7777/api/health | jq -r '.secrets.db_password')
+          if [ "$PROV" != "generated" ]; then
+            echo "::error::secrets.db_password should still be 'generated' after recreate, got: $PROV"
+            docker compose logs pinchy | grep -i "db-password" || true
+            exit 1
+          fi
+          echo "Generated DB password survived the recreate (steady state)"
 
       - name: Verify Pinchy reconnects to OpenClaw after recreate
         run: |
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -297,6 +297,7 @@ jobs:
     # a .env file.
     env:
       PINCHY_VERSION: latest
+      DB_PASSWORD: pinchy-ci-smoke-pass
 
     steps:
       - uses: actions/checkout@v4
@@ -1062,6 +1063,7 @@ jobs:
     # in docker-compose.e2e.yml disables that limit only for this E2E stack.
     env:
       PINCHY_VERSION: latest
+      DB_PASSWORD: pinchy-ci-smoke-pass
 
     steps:
       - uses: actions/checkout@v4
@@ -1182,6 +1184,7 @@ jobs:
 
     env:
       PINCHY_VERSION: latest
+      DB_PASSWORD: pinchy-ci-smoke-pass
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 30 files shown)


clemenshelm:
Update after the first CI round: the fail-closed check killed all 7 CI stacks that boot the production image on the default password — i.e. it worked exactly as designed, just also on our own test stacks. Second commit fixes that:

- the 7 production-image stack jobs in `ci.yml` now set `DB_PASSWORD` next to their existing `PINCHY_VERSION` env (fresh stacks initialize Postgres with it),
- `getting-started.mdx` got the `.env` step it was already missing — the documented quick start failed on `${PINCHY_VERSION:?}` even before this PR, since it never wrote a `.env`. The new step generates `PINCHY_VERSION` + `DB_PASSWORD` in one block, matching what the end-user install simulation already does,
- `installation.mdx` now documents `DB_PASSWORD` as required in production.

The end-user install/upgrade simulations were green throughout — they already generate their own `DB_PASSWORD`, which is good evidence the canonical install path is unaffected.

clemenshelm:
**Direction change after discussion: no more breaking change.** The fail-closed startup check is replaced by a boot-time auto-migration:

- `entrypoint.sh` now runs `scripts/resolve-db-password.mjs` before drizzle-kit and the server. Installs that never set `DB_PASSWORD` get a generated credential on their first boot after the upgrade — persisted in the `pinchy-secrets` volume (the `getOrCreateSecret` pattern), applied via `ALTER USER`, invisible to the operator. Persist-before-alter + a recovery path make every crash point repeatable.
- An explicit `DB_PASSWORD` always wins — including the previously broken case where someone set `.env` without running `ALTER USER` manually; the resolver now heals that too.
- If migration is impossible (db down, volume read-only), the server warns loudly and runs — never refuses to boot.
- Provenance gains `generated` (Settings: *Auto-generated (Docker volume)*), and the preload now marks file-loaded `BETTER_AUTH_SECRET` so the Secrets card stops mislabeling it as an env var.
- The **docker-smoke CI job now boots without `DB_PASSWORD`** and asserts `secrets.db_password == "generated"` plus persistence across its down/up recreate — a permanent E2E regression test of the migration. The six E2E stacks keep their explicit password (operator path).
- Docs: quick start is zero-config again (`PINCHY_VERSION` only), the upgrade section describes the automatic migration, no breaking-change recipe needed.

Test evidence: 13 unit tests (state machine incl. crash ordering), 5 real-Postgres integration tests (ALTER USER round trip, idempotency, crash recovery, operator override incl. stale-file cleanup, SQL metacharacter quoting), and a full local production-image compose stack: fresh boot migrates → restart stays steady → down/up keeps the credential → late `DB_PASSWORD` heals to custom.

Issue #124, opened 2026-04-13
Approval Queue: interactive, role-aware human-in-the-loop approvals


clemenshelm:
## Problem
Pinchy currently has no interactive, runtime approval queue for sensitive tool actions. In multi-user setups, this is a blocker: users either get too much power or workflows are blocked entirely.

Dominikus' real-world test highlighted this gap: he needed human approval checkpoints for orchestration flows, but had to fall back to hard permissions.

## Goal
Introduce a built-in **Approval Queue** that supports admin/user roles and action-level decisions.

## Scope
- Queue pending approvals for selected high-risk tool calls/actions
- Role-aware approvals:
  - regular users can request approvals
  - admins (or configured approvers) can approve/deny
- Approval record in audit trail (who approved what, when, why)
- Timeout/expiry behavior for pending approvals
- Optional notification hooks (future: Slack/Email)

## Acceptance Criteria
- [ ] Admin can configure which tools/actions require approval
- [ ] Agent pauses execution and creates an approval request when required
- [ ] Approver can approve or deny from Pinchy UI
- [ ] On approve: action executes; on deny: action is blocked with explicit reason surfaced
- [ ] Audit trail logs full approval lifecycle
- [ ] Works with multi-user RBAC/groups

## Why this matters
- Core enterprise requirement (governance + accountability)
- Aligns with Pinchy philosophy: safe defaults, controlled power
- Enables orchestrator use cases without granting full shell/raw access

rpelevin:
I would make the approval request identity the spine of this feature.

The queue should not just say "this tool needs approval"; it should bind one approver decision to one pending action.

A useful approval request record could include:

- approval request id;
- run/session id;
- requesting user id;
- agent or actor id;
- tool/action name;
- target resource;
- canonical args or payload digest;
- policy id and policy version;
- risk class;
- eligible approver set;
- decision: pending, approved, denied, expired;
- decided_by, decided_at, and decision reason;
- resume token or execution correlation id.

Tests I would add:

1. approval resumes exactly the pending action that created the queue item.
2. changing tool, target resource, or args digest requires a new approval.
3. a requester cannot approve their own request unless an explicit admin policy allows it.
4. expired approval requests fail closed and do not execute later.
5. duplicate approve/deny clicks are idempotent.
6. the audit trail orders request, decision, and execution as separate records.

That keeps the queue from becoming a broad role grant while still making the UX simple.

clemenshelm:
[...]

PR #517, opened 2026-06-16
feat(approvals): human-in-the-loop tool-call confirmation (#124 Tier 1+2)

clemenshelm:
## What

Implements **Tier 1+2 of #124** — human-in-the-loop tool-call confirmation. An admin marks tools as "require confirmation"; the agent then **pauses and asks the acting user** before running them. The user approves their own action in seconds — no admin bottleneck.

This follows the reframe in #124: the validated, market-standard need is *confirmation* (the acting user confirms, so the agent can't act autonomously), not an admin approval queue. Four-eyes (a *different* approver) stays documented in the issue as a demand-pulled extension and is **not** built here.

## How

A two-layer, **server-enforced** design:

- **Gate** — a new internal hook-only plugin `pinchy-approvals` registers a `before_tool_call` hook that asks Pinchy's `gate-check` endpoint for every tool. The endpoint is the security boundary; it fails **closed** if unreachable.
- **Durable ticket** — Pinchy's DB owns the lifecycle. A grant is bound to `(agentId, requesting user / senderId, args digest, originating session)` and **consumed exactly once** (`FOR UPDATE SKIP LOCKED`). Changed args ⇒ new confirmation. Never shared across users of a shared agent. Pending confirmations expire (15 min) and fail closed.
- **Resolve** — the acting user approves/denies their own request (self-confirm enforced server-side); the agent re-issues and the gate consumes the now-approved ticket.

## Surfaces

- **Admin config**: a "Require confirmation before" section in the agent permissions tab, with per-tool checkboxes and a "Use recommended" button (auto-selects the agent's `powerful` tools).
- **Approve/deny**: an in-app `ApprovalsInbox` that lists the user's pending confirmations.
- **Audit**: full `approval.requested → granted/denied → consumed/expired` lifecycle. PII-free — only the args **digest** is logged; the human-readable summary stays on the operational row.
- **Docs**: a "Require Tool Confirmation" guide.

## Tested

- **Unit**: 5992 passing (digest, gate decision, policy, plugin gate logic, two RTL components).
- **Integration (real PostgreSQL)**: 133 passing — consume-once / fail-closed / concurrent-consume on real SQL; gate-check, list, and self-confirm decision routes with their audit lifecycle.
- **E2E**: a gated tool is blocked → `approval.requested` → user grants → `approval.granted` (runs in the integration E2E CI job; satisfies the internal-plugin coverage contract).
- `pnpm build`, `tsc --noEmit`, `eslint` (0 errors), and the docs build all pass. All Plugin-Integration-Contract drift guards updated.

## Deferred (documented, not built)

- **Tier 3 four-eyes** admin queue — demand-pulled (see #124).
- Native `requireApproval` **inline-in-chat** card + automatic proceed-injection (would also unlock Slack/Telegram approval routing) — needs an `openclaw-node` extension. Interim UX: approve in the inbox, then ask the agent to proceed.

Refs #124.


clemenshelm:
Superseded by #865, which relands this work on current main.

This branch was harvested rather than rewritten: all 50 files here are present in #865. The only differences are the migration (renumbered several times as main moved, now `0068`, SQL byte-identical) and seven files added by review — the expiry/retention sweep, the shutdown wiring, and their tests.

#865 additionally carries the fixes that came out of reviewing and actually running it: grants bound to the tool name as well as the argument digest (two tools with identical arguments could otherwise share one approval), expiry honoured in the inbox and the decision route, the `pinchy-approvals` mount the dev stack was missing, non-registry tools offered for confirmation, and a block reason that no longer leads the model to invent a `/approve` command.

Closing to keep one open PR for this feature. The branch `feat/approval-confirm` stays as-is, and this can be reopened if ever needed.

Refs #124, #859.

Issue #483, opened 2026-06-11
feat(usage): lossless per-turn token accounting instead of gauge sampling (target v0.6.0)


clemenshelm:
**Target: v0.6.0.** Follow-up to #482, which fixed the cache-token field mapping but left the accounting fundamentally lossy.

## Problem
The usage poller samples OpenClaw's per-session token counters on an interval. Those counters are **last-turn gauges** — OC overwrites them with each turn's values (verified live on staging, OC 2026.5.28; also the #426 lesson). Sampling a gauge cannot reconstruct totals:

- Turns completing between two polls are simply lost (observed: 4 snapshots captured of a 9-turn session).
- The watermark-delta logic in `recordUsage` assumes **cumulative** counters; against last-turn gauges its clamped deltas produce noise (verified: stored values decode exactly as `max(0, turnN_value − turnM_value)` between unrelated turns).
- Result: day totals are wrong in both directions and `estimated_cost_usd` inherits the error.

## Proposed direction
Account **per turn** from data OC already persists exactly:

- The trajectory JSONL has one `model.completed` event per turn with precise usage (`input`, `output`, `cacheRead`, `cacheWrite`, `total`) — the diagnostics exporter already parses these (`turn-extractor.ts`).
- Options, roughly in order of preference:
  1. **Consume usage from the chat stream**: openclaw-node `ChatChunk`s carry `runId`; if OC emits usage on the final chunk (or via an event), record per-turn usage at stream end in `client-router.ts` — no polling, no files. Check what OC exposes; we own openclaw-node and can extend it (e.g. surface usage events) if the gateway provides them.
  2. **Tail/scan trajectory files** for `model.completed` events (watermark by `seq` per session) — exact, but couples Pinchy to OC's on-disk format.
  3. Upstream: ask OC for cumulative session counters or a usage API; sampling becomes correct once counters are cumulative.
- Dedupe by `(sessionId, runId)` so restarts/re-reads never double-count.
- Keep the poller as fallback for sessions not driven through Pinchy's chat path (cron, channels) until the chosen mechanism covers them.

## Acceptance
- A multi-turn session records one usage row per turn (or an exact per-turn aggregate), matching the trajectory's `model.completed` usage sums exactly.
- Dashboard day totals equal the sum of trajectory usage for that day.
- Cost estimates use the exact per-turn token classes (uncached input, cache read, cache write, output).
[...]

Issue #314, opened 2026-05-09
regenerateOpenClawConfig occasionally drops Telegram state, triggering gateway restart cascade (#311 root cause)


clemenshelm:
## Summary

`regenerateOpenClawConfig()` intermittently produces a payload that is missing `channels.telegram`, `bindings`, and `session.identityLinks` even when the DB has the corresponding rows. The resulting tiny payload is written to disk via `writeConfigAtomic`, OpenClaw's inotify watcher diffs it against its in-memory baseline, and triggers a full gateway restart cascade. The same payload is then sent via `client.config.apply` and rejected by OpenClaw's own size-drop validation (`size-drop:OLD->NEW`).

This is the root cause of the long-standing `e2e/telegram/agent-create-no-restart.spec.ts:216` flake referenced as "the known flake" in `.github/workflows/ci.yml:1247-1260`.

## Defense-in-depth landed in #311

PR #311 added a Pinchy-side size-drop guard (`build.ts` ≈ line 980) that mirrors OpenClaw's threshold (>50%) BEFORE `writeConfigAtomic` runs. When the guard fires, the bad payload is saved to `${CONFIG_PATH}.regenerate-rejected.<ts>` for postmortem and the regenerate returns without writing. This prevents the cascade but does **not** fix the underlying race — it converts a silent corruption into a loud, observable failure.

## Evidence (CI run 25596470147 / job 75143153379)

- **Failure**: `e2e/telegram/agent-create-no-restart.spec.ts` ✘
- **OpenClaw log**:
  - \`[reload] config change detected; evaluating reload (gateway.controlUi.enabled, discovery, update, canvasHost, secrets, agents, plugins, models, channels, bindings, session)\` — basically every section diff
  - \`Config write rejected: /root/.openclaw/openclaw.json (size-drop:4874->2351)\`
  - \`config.apply 33228ms errorCode=UNAVAILABLE\`
- **DB state at diagnostic time**: telegram_bot_token present for Smithers (`facb467d-...`)
- **File state at diagnostic time**: full Telegram block + bindings + session — meaning the file was healthy AFTER the failure, but Pinchy's regenerate at the failure moment produced 50% smaller content

## Reproducer (hypothesized)

The test sequence:
1. `connectBot(smithersAgentId, BOT_TOKEN)` — writes `telegram_bot_token:...` to DB and uses targeted-write to update `channels.telegram` in the file
2. \`waitForOpenClawQuiet()\` — 30s of stable state
[...]

PR #533, opened 2026-06-17
fix(usage): retry transient EACCES when reading OpenClaw session files

clemenshelm:
## Why

The **Integration Tests (OpenClaw + Fake Ollama)** job is intermittently red on `main`. Pulling the real CI logs across four failed runs showed the recurring main-wide failure is `usage-tracking.spec.ts:106` — `expect(delta.input).toBe(42 * delta.rows)` failing with `delta.input` = 84 (post-`0fc858ee`) or 18128 (pre-`0fc858ee`), passing on re-run.

Root cause is a real **production data-integrity race**, not just a test issue:

- Per-turn usage recording (`recordSessionTurnsUsage`, the chat-`done` trigger + the poller backstop) reads `sessions.json` and `*.trajectory.jsonl` from the shared `openclaw-config` volume.
- OpenClaw (root) rewrites those files as mode `0600` on every session update. `start-openclaw.sh`'s ~50ms chmod loop reopens them to `0644/0755`, but Pinchy (uid 999, in prod and in CI) can read in that window and hit a **transient `EACCES`**.
- `resolveSessionId` / `readTrajectoryJsonl` only handled `ENOENT`, so a transient `EACCES` aborted the read and the turn's usage was **silently dropped** until a later poll.
  - In production: usage is under-counted on the permission race.
  - In CI: the delayed row lands in a *later* `usage-tracking` spec's `before→after` window instead of its own baseline — contaminating the exact-token assertion (e.g. a preceding `pinchy_read` tool turn's row showing up as `delta.input=84`).

## What

Add a bounded `EACCES` retry to both readers in `jsonl-reader.ts`, mirroring the established `#314` retry pattern in `openclaw-config/write.ts` (`5 × 100ms` covers two chmod-loop ticks), async since these readers are already async:

- Transient `EACCES` → retry after a chmod-loop tick.
- `ENOENT` → unchanged ("no session/trajectory yet", poller backstops).
- Persistent `EACCES` → still propagates (logged) so a real permission problem isn't masked.

Recording becomes prompt, so in the serial integration suite each turn's row lands in its own measurement window rather than a later spec's.

## Explicitly NOT touched (no rollback of recent work)

- `#483` per-turn accounting + `(sessionKey, runId)` dedup — unchanged.
- `0fc858ee` fake-ollama deterministic-usage paths — unchanged.
- Authoritative liveness state machine, removed orphan/watchdog timers, the deliberately-dropped seq-keyed `session-projection.ts` — untouched.
- The `start-openclaw.sh` chmod loop (incl. its ctime-churn guard) — untouched; this is a reader-only change.

## Tests

- New `jsonl-reader-eacces-retry.test.ts`: transient `EACCES` retried and resolves; persistent `EACCES` propagates within a bounded budget; `ENOENT` not retried (returns null / throws `TrajectoryFileNotFoundError` on first miss). Written test-first (RED → GREEN).
- Existing `jsonl-reader.test.ts` and the usage unit suites (`usage-per-turn`, `usage-poller`, `usage-from-trajectory`) still green.

## Note

This fixes the dominant, main-wide usage-tracking flake. A separate flake in `18-chat-liveness.spec.ts:155` (mid-stream reload dropping the first streamed word) is a distinct resume-path root cause and will be addressed in its own focused PR.

### DIFF
--- a/packages/web/src/__tests__/lib/diagnostics/jsonl-reader-eacces-retry.test.ts
+++ b/packages/web/src/__tests__/lib/diagnostics/jsonl-reader-eacces-retry.test.ts
@@ -1,0 +1,86 @@
+import { describe, it, expect, vi, beforeEach } from "vitest";
+
+// OpenClaw (root) rewrites `sessions.json` / `*.trajectory.jsonl` as mode 0600
+// on every session update; start-openclaw.sh's chmod loop reopens them to
+// 0644/0755 within ~50ms, but Pinchy (uid 999) can read in that window and hit
+// a TRANSIENT EACCES. The per-turn usage recorder reads these files on the
+// low-latency chat-`done` path and the poller backstop — if a transient EACCES
+// silently aborts the read, the turn's usage is dropped until a later poll,
+// which under-records usage in production and (in CI) leaks a prior turn's row
+// into a later test's usage-tracking measurement window.
+//
+// We can't reproduce a privileged chmod race against real files in a unit test,
+// so we mock `readFile` to drive the transient-EACCES sequence deterministically
+// and assert the retry contract (mirrors the established #314 retry pattern in
+// openclaw-config/write.ts: a bounded budget that covers two chmod-loop ticks).
+const { readFileMock } = vi.hoisted(() => ({ readFileMock: vi.fn() }));
+vi.mock("node:fs/promises", async (importOriginal) => {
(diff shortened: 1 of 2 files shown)


clemenshelm:
Companion fix for the same integration-job flakiness: #535 addresses the second main-wide flake (`18-chat-liveness:155` mid-stream-reload prefix loss). Together they cover both flakes; this PR is the dominant one (usage-tracking).

Issue #310, opened 2026-05-09
UX: auto-refresh chat history when WS reconnects (no page reload needed)


clemenshelm:
## Problem

When Pinchy restarts mid-turn, the browser briefly shows a "Connection Lost" error. After reconnect, the in-progress assistant reply disappears from the UI (it was in React state, not yet persisted to history). The user must do a manual page reload (Cmd-R) to see the full conversation.

The messages **are** persisted in OpenClaw JSONL (Layer B / #297 ensures this). The UX gap is that the client doesn't re-fetch history automatically on reconnect.

## Expected behaviour

Within a few seconds of Pinchy reconnecting (WS `CONNECTED` frame received), the chat history should auto-reload — user message and full assistant reply become visible without a manual page reload.

## Relevant code

- `packages/web/src/hooks/use-ws-runtime.ts` — `onclose` injects a synthetic disconnect-error message; the `CONNECTED` frame handler already sets `setIsHistoryLoaded(false)` / `setKnownEmptyHistory(false)`. May not reliably trigger a new `history` WS message.
- The frontend sends `{ type: "history", agentId }` when mounting or when `isHistoryLoaded` becomes false.

## Subtleties

- The synthetic error message should be replaced by real history on reload (not shown alongside it).
- Avoid triggering auto-refresh if the user is mid-typing (non-empty input field).
- Avoid a flicker loop if reconnects are frequent.

## Scope

Small UX improvement — no new infrastructure needed, just wire the existing reconnect path to re-request history. TDD: test in `packages/web/src/__tests__/` first.

Part of the #199 follow-up (Layer A closure).

clemenshelm:
## Root cause precisely located + research on how other tools solve this

Spent some time tracing the production failure on `pinchy.heypinchy.com`. Two specific agents (Penny on Bookkeeper template, Piper on CRM template) consistently trigger this. We have a smoking-gun trace, a one-line client fix candidate, and three tiers of mitigation aligned with current state-of-the-art patterns.

### TL;DR

[...]

PR #555, opened 2026-06-19
feat(chat): durable, honest agent-error visibility (paused banner)

clemenshelm:
## Problem

In production, an agent hit a provider **rate limit** mid-run (after creating Odoo draft bills). The audit log recorded `chat.agent_error` (`errorClass: "transient"`), but the chat showed **no feedback**.

Root cause: the live error bubble is ephemeral client state. It's never persisted, so a reload — or a WS reconnect during a long tool loop — loses it, leaving only the audit row. Pinchy has no chat-message persistence of its own (OpenClaw owns history, and errored turns are deliberately not written there).

## Approach — a durable, server-backed "paused" banner

Rather than inject persisted errors into the WS history (which would require rewriting the fragile message-reconcile gate — the #1 risk an adversarial design review flagged, with a documented #310/#199 duplicate-bubble history), the durable error is a **separate, server-backed surface**: a banner above the composer fed by a small endpoint. This sidesteps the reconcile path entirely and the messageId-placement problem.

## What's in here (7 commits)

1. **Honest, cause-specific copy** — a `transientError` chat-frame payload + a dedicated `ChatErrorMessage` branch. The `transient` audit class spans rate-limit / overloaded / timeout / 529, so the bubble names the **actual** cause instead of always saying "rate limit". When the failed run already executed a tool (`sideEffects`), it warns that retrying may **duplicate** already-applied writes. Historical (reloaded) errors use `role=status`; live ones stay `role=alert`.
2. **Durable store** — `chat_session_errors` table + module. The latest un-superseded, un-dismissed row for an **exact** sessionKey is the active error. Supersede is scoped to the triggering `clientMessageId` (a later unrelated message succeeding must NOT clear an unanswered error); dismiss is owner-scoped.
3. **`directSessionKey`** — single source of truth for the session key, shared by the WS router and the new route.
4. **Stream-loop wiring** — on an error chunk (and the silent-stream synthesis) mirror the error into the store next to the existing audit; on a successful `done`, supersede; a `tool_use` chunk flips `sideEffects`. All best-effort, never breaks the stream.
5. **Endpoints** — `GET /active-error` (caller-scoped, optional `?chatId`) and an owner-scoped `DELETE` dismiss that writes a `chat.error_dismissed` audit row.
6. **Banner** — `ChatErrorBanner` fetches on mount and re-surfaces the error; Dismiss clears it server-side; Retry resends the last user message, gated behind a duplicate-write confirmation when `sideEffects`.
7. **Retention sweep** — reaps resolved rows past 30 days on the upload-GC cadence, one summary audit row per `sweepId`.

## Why a dedicated table (not audit reuse)

Audit rows are append-only + per-row HMAC — a supersede/dismiss `UPDATE` would break `verifyIntegrity`; audit detail is PII-scrubbed/truncated and lacks the `clientMessageId`/`runId`/`sideEffects` granularity; and chat loads shouldn't query the compliance table on the hot path.

## Tests

- Unit: transient classifier + `ChatErrorMessage` branch (copy by cause, sideEffects warning, role, dismiss); `ChatErrorBanner` (fetch/dismiss/retry/confirm); `directSessionKey`.
- `client-router` (153): error chunk persists with `sideEffects`/`transientReason`/`clientMessageId`; `done` supersedes the triggering message.
- Real-DB `test:db`: store record/active/supersede-by-message/owner-dismiss/exact-scoping/newest; GC retention + sweepId audit.
- Full unit suite (6225) + full `test:db` (123) + `tsc` + ESLint + Prettier all green.

Wire format (`tool_use` / `error` chunk types) verified against the real `openclaw-node@0.13.0` `ChatChunk` type, not invented.

## Remaining verification

An **E2E** (fake-ollama rate-limit injection → banner → reload → still there → dismiss → gone, plus reconnect-mid-run) is the one piece left. fake-ollama can't emit a provider error today, and OpenClaw's error translation is third-party, so that spec needs to be authored **against the running Docker stack** to avoid a false-green/flaky test — tracked as a follow-up rather than shipped blind.

---

Follows **PR #548** (audit-sidebar UX). Both stem from the same production observation; independent, not stacked.


clemenshelm:
## E2E added ✅ — and it earned its keep

The "remaining verification" E2E is now in (`e2e/integration/19-durable-agent-error-banner.spec.ts` + two fake-ollama error-injection triggers). It drives the whole path through the **real OpenClaw container**: fake-ollama returns a 429 → OpenClaw surfaces a transient error chunk → Pinchy persists it → after a **reload** the banner re-fetches and re-surfaces it → dismiss clears it; plus a tool-then-429 case asserting the duplicate-write warning + confirm-gated retry.

Run red→green against the Docker stack (`docker-compose.yml` + `e2e.yml` + `integration.yml`), it caught **two real OpenClaw-integration bugs the unit tests missed** — both because the unit tests restated a chunk shape OpenClaw doesn't actually produce:

1. **Supersede-on-errored-`done`.** OpenClaw emits a terminal `done` *after* the `error` chunk, so the supersede fired on every `done` and immediately cleared the error — the banner never appeared. Fixed by gating supersede on a genuinely successful turn.
2. **`sideEffects` never set.** OpenClaw doesn't surface tool execution as a chat chunk (no `tool_use`), so the chunk-based detection always read false. Now derived from the audit trail (`tool.*` row scoped by `resource = agent:<id>` since the run began), on the rare error-persist path only.

Both green now (2 passed, ~57s). The new spec runs in the existing integration-e2e CI job automatically.

Issue #556, opened 2026-06-19
Tracking: use OpenClaw more, fight it less — integration audit (2026-06-19)


clemenshelm:
## Context

Audit of where Pinchy under-uses OpenClaw, reimplements things OpenClaw already does, or works against it. Source: openclaw-node 0.13.0 + OpenClaw 2026.6.5 surface vs. Pinchy `main` (post-v0.6.0). This is the umbrella so the individual findings stay connected — pick work off the table below.

Each row was verified against `origin/main` (not the stale working copy) and against the actual openclaw-node API. Where a "deliberate divergence" existed, the original reason was re-checked for staleness.

## Findings → issues

| Finding | Issue | Priority | State of play |
|---|---|---|---|
| **Re-enable the chat stop button** — `chatAbort` against `agent` RPC was a no-op (openclaw#42172); now fixed upstream and shipped in our pinned OC 2026.6.5. | #550 | **P1** | Blocker gone. Needs empirical Step-0 verification before re-wiring. Was shipped (PR #23) + rolled back (PR #136) once. |
| **Audit coverage for OC built-in tools / subagents / lifecycle** — `pinchy-audit` plugin already captures tool calls via hooks, but no E2E proof for built-ins, subagent spawns unverified, lifecycle events unaudited. | #551 | **P1** (eu-ai-act) | Plugin exists & always-on. Gaps are verification + subagent/lifecycle. |
| **`pairingRequired` event handler** — openclaw-node emits it; Pinchy swallows the 1008 close → silent reconnect loop, no audit row. | #164 | **P1** | Smoke test pins the API; no production handler. Phase 1 is ~3-4h additive work. |
| **Auto-compact at token threshold** — manual compaction shipped (#448/#536); auto-trigger missing. | #547 | Backlog (needs design) | 6 open design questions: context-size exposure, model-window resolution, audit semantics, interaction with OC native compaction (commit 57790c234). |
| **Telegram pairing via `pairing.*` RPC** instead of shared-volume file reads. | #549 | Backlog (deferred) | RPCs exist. Memory's reason (channel-restart bug openclaw#47458) is **still valid** in 2026.6.5 — but orthogonal to this migration. Sequence after #476/#477. |
| **usage-poller idle backoff** (not event-push — OC emits no usage event). | comment on #261 (sub-task D) | Backlog | Per-session backoff is the right call; event-sub is a dead end. Implementation outline posted. |
[...]

Issue #508, opened 2026-06-16
Chats: per-agent session overview + per-task session model (fixes silent history loss from cross-channel /new)


clemenshelm:
## Summary

Add a per-agent **"Chats"** overview (header chat-switcher) so users can see and return to their
past conversations with an agent, and adopt OpenClaw's native **per-task / explicit-key** session
model. This is the v1 **safety net**: no conversation should silently become unreachable again.

It also **fixes a real production data-loss footgun**: an accidental `/new` sent from **Telegram**
(an OpenClaw gateway slash command → `sessions.reset(reason:"new")`) reset the session that — via
`identityLinks` + `dmScope:"per-peer"` — was **shared** between Telegram and the web UI, so the web
chat history became unreachable (the old transcript stays on disk). It was confirmed **not** the
4 AM daily-reset (the `session.reset = {mode:"idle", idleMinutes:525600}` override is deployed and
effective in OpenClaw 2026.5.28).

The deeper gap: Pinchy can already *create* sessions (involuntarily via Telegram `/new`, deliberately
via #451) but offers **no way to see or return to** past sessions.

## OpenClaw model (verified against types + docs + prod)

- **Session** = durable context + transcript (`sessionId` → `<sessionId>.jsonl`), addressed by an
  explicit `sessionKey`. One active session per bucket.
- **Bucket** for DMs = `dmScope` (`main` default, `per-peer`, `per-channel-peer`, …). There is **no
  native "multiple parallel chats per direct user"**; multiple buckets only via explicit keys.
- **Conversation** = channel-side thread bound to a session; `identityLinks` unifies channels onto one
  session. `topic`/`thread` (+ native `AutoTopicLabelConfig` LLM naming) are **channel-coupled**, not a
  web multi-chat primitive.
- **Reset** archives the old transcript on disk (retained per `resetArchiveRetentionMs`, then GC'd).
- **RPCs:** `sessions.create` (explicit `key`, `label`/`name`, `parentSessionKey`, optional initial
  message; used by OC's own webchat), `sessions.list({kinds})`, `sessions.patch` (label),
  `sessions.history`, `sessions.subscribe`/`sessions.changed`. **No public resume/fork RPC** for an
  orphaned transcript.
- **Core tension:** you cannot natively have BOTH forced cross-channel unification AND multiple parallel
  web chats → embrace the **per-task** model (the direction already noted in #277).

## v1 design decisions

[...]

PR #563, opened 2026-06-19
feat(chat): re-enable the stop button via chatAbort (#550)

### LINKED ISSUE #550: Re-enable the chat stop button (upstream openclaw#42172 is fixed)
## History — this has been tried once and rolled back

This isn't a fresh feature. We shipped it in **PR #23** (March 2026) and rolled it back in **PR #136** (April 2026). #20 closed as "blocked on upstream".

What went wrong:

- PR #23 wired `ComposerPrimitive.Cancel` → WS abort frame → `client-router.ts` → `openclawClient.chatAbort(sessionKey)`.
- The button looked like it worked client-side (UI flipped, partial response preserved, composer re-enabled).
- But on the gateway side, **`chat.abort` returned `{ aborted: false }`** for every run Pinchy ever dispatched.
- Root cause: Pinchy's `openclawClient.chat()` goes through the `agent` gateway RPC, not `chat.send`. OpenClaw's `chat.abort` handler only looked up abort controllers registered by `chat.send`. The `agent` RPC handler (`dispatchAgentRunFromGateway`) never created one, so abort silently no-op'd. Tracked as [openclaw/openclaw#42172](https://github.com/openclaw/openclaw/issues/42172).
- Side effect: subsequent `chat()` calls on the same session returned empty responses — the gateway's `HELD_LOCKS` / `ACTIVE_EMBEDDED_RUNS` weren't released for unaborted-but-disconnected runs. Three workaround attempts (client reconnect, server stream drain, server stream sync) all failed because the lock lived inside the gateway.
- PR #136 removed the button rather than ship a placebo.

## Why now (upstream fix landed and shipped)

- **openclaw/openclaw#42172 is closed** (2026-04-26).
- Upstream commit `fd68994c5a` (2026-03-14, `src/gateway/server-methods/agent.ts`) creates the `AbortController` before `agentCommandFromIngress`, registers it in `context.chatAbortControllers`, passes `abortSignal` down, and cleans up in `.finally()`.
- Pinchy currently pins **OpenClaw 2026.6.5** (released 2026-06-09) — three months after the fix shipped.
- `openclaw-node` 0.13.0's `chatAbort(sessionKey, runId?)` is correct: it both kills the local generator and dispatches the `chat.abort` RPC. The `chat.test.ts` suite covers the abort path including the local-generator-termination case.

So the original blocker is gone. We can re-introduce the button, but **the first step is empirical confirmation** — we got bit last time because the client-side flow lied to us about the server having stopped.

## What it means in code today

Same gap as before — the rollback removed everything:

- `packages/web/src/components/assistant-ui/thread.tsx:415-426` — `ComposerPrimitive.Cancel` button is rendered while `isRunning`.
- `packages/web/src/hooks/use-ws-runtime.ts:1856-1865` — `useExternalStoreRuntime({ messages, isRunning, convertMessage, onNew, adapters })` has no `onCancel`.
- `packages/web/src/server/client-router.ts:104-111` — `BrowserMessage = ChatMessage | HistoryRequestMessage`, no abort variant.
- `packages/web/server.ts:381` + `packages/web/src/server/run-watchdog.ts:143` — `chatAbort` is wired into the server-side stuck-run watchdog and confirmed working in the auto-recovery path.

### Step 0 — empirical verification (must succeed before any other work)

Add an integration test against a real OpenClaw 2026.6.5 container (the existing `docker-compose.integration.yml` stack) that:

1. Dispatches a chat via the `agent` RPC (Pinchy's standard `openclawClient.chat()` path), with a fake-ollama trigger that streams slowly for ≥10 s.
2. Mid-stream calls `openclawClient.chatAbort(sessionKey, runId)` directly.
3. Asserts the `chat.abort` response is `{ aborted: true }` — **not** `{ aborted: false }` as in March.
4. Immediately dispatches a second `chat()` on the same `sessionKey` and asserts it gets a real reply, **not** an empty `done` chunk. This is the session-lock regression test.

If Step 0 passes, proceed. If it fails — even partially — we don't ship anything user-facing. Drop a comment here, leave the button off, file a follow-up against the next OC release.

### Step 1 — re-wire (mostly restoring PR #23, with diffs informed by the rollback PR #136)

The discriminated union and message-routing scaffolding from #23 were preserved in #136 ("clean refactor worth preserving"). We're mostly adding back what #136 removed:

1. Add `AbortMessage = { type: "abort", clientMessageId?: string }` to `BrowserMessage` in `client-router.ts`, plus a Zod schema next to the other message schemas (#508 added that convention).
2. Pass `onCancel: () => sendAbort()` into `useExternalStoreRuntime`. Resolve the active `runId` from the run-registry that #441 added (`packages/web/src/server/run-watchdog.ts` already tracks it).
3. In `ClientRouter.handleMessage`, route the abort message to `openclawClient.chatAbort(sessionKey, runId)`.
4. **Audit event** `chat.user_aborted` with `{ chatId, sessionKey, runId, outcome }`. Distinguishes user-initiated aborts from the watchdog-driven `chat.run_no_first_chunk` we already audit. Enterprise-relevant because "did the user actually cancel that browser run" is a real CISO question.

### Step 2 — explicit-abort touch points

The same abort-the-current-run primitive is needed by:

- `ChatSwitcher.startNewChat()` and any other chat-unmount path (today the chat unmount severs the local stream but leaves the OC run going — see #451 close-comment).
- Page-navigation away from a streaming chat.

Wire these to the same `onCancel`.

### Step 3 — design choice from #23: composer always active

PR #23 also moved the stop button *out* of the composer onto the streaming assistant message and kept the composer enabled while streaming (the ChatGPT pattern). That's a separate UX call — could land in this PR or be deferred. Recommend deferring to a follow-up so this one stays scoped to the regression-fix part.

## Acceptance criteria

- [ ] Integration test (Step 0) green against OC 2026.6.5: `chatAbort` against an `agent`-RPC run returns `{ aborted: true }`, the session is reusable immediately after.
- [ ] Clicking stop while streaming dispatches the abort frame and tokens-after-abort (verified via `usage-poller` per-turn accounting from #483) reflect the truncated run, not the full would-have-been run.
- [ ] `audit.event_type = "chat.user_aborted"` with `outcome: "success"` on RPC ack.
- [ ] Existing watchdog-driven aborts keep working (regression test on `run-watchdog.test.ts`).
- [ ] E2E with fake-ollama slow-stream: dispatch, cancel, then dispatch a second message on the same chat — both succeed.
- [ ] Manual smoke on staging before merge: cancel a real browser tool call, then send a follow-up. Both behave normally.

## References

- #20 (closed) — original bug report, root-cause comments
- PR #23 (closed) — initial implementation
- PR #136 — rollback with explanation
- [openclaw/openclaw#42172](https://github.com/openclaw/openclaw/issues/42172) — upstream fix (closed 2026-04-26, shipped via commit fd68994c5a)
- PR #441 / `run-watchdog` — gives us the run-registry we need for `runId` lookup
- #550's prior incarnation (this issue's original body) referenced openclaw#42172 implicitly via the "today it does nothing meaningful" framing — now we have the history.

clemenshelm:
Closes #550. Part of the OpenClaw-integration audit (#556).

## Why

The chat composer's stop button has been cosmetic since **PR #136** rolled it back. The original implementation (**PR #23**) looked like it worked client-side, but OpenClaw's `chat.abort` was a **no-op on `agent`-RPC runs** — it returned `{ aborted: false }` and never released the session lane, so every subsequent message on that session came back as an empty `done`. Root-caused as [openclaw/openclaw#42172](https://github.com/openclaw/openclaw/issues/42172).

That upstream bug is **fixed** (closed 2026-04-26, commit `fd68994c5a` registers the `AbortController` for agent-RPC runs) and shipped in our pinned **OpenClaw 2026.6.5**. So the button can come back — but only with empirical proof against a real gateway, because last time the client-side flow lied about the server having stopped.

## What changed

- **Server** (`client-router.ts`): a new `{type:"abort"}` WS frame is routed to `chatAbort(sessionKey, runId)` — runId resolved precisely from the active-run registry (never aborts a newer run that already replaced this one), falling back to a sessionKey-only abort. It emits **`chat.run_aborted`** (reserved by #441's watchdog work, **emitted here for the first time**) with `outcome` reflecting the RPC result, so an operator always sees that a user asked to stop. The abort frame goes through the same agent-access check as chat/history.
- **Client** (`use-ws-runtime.ts`): `onCancel` wired into the assistant-ui external-store runtime — sends the abort frame and optimistically ends the turn so the composer re-enables immediately. The existing `ComposerPrimitive.Cancel` button now does something; no markup change needed.
- **Docs**: `chat.run_aborted` flipped from *reserved* to *emitted* in `audit-trail.mdx`.

The stream pipe's existing post-abort handling (the synthetic terminal `done` openclaw-node emits) cleans up the run — no new teardown code.

## Test plan

- **Unit** — `client-router-abort.test.ts` (6): routing to `chatAbort` with the registry runId, the `chat.run_aborted` audit (actor=user, success), sessionKey-only abort when no run is registered, per-chat `chatId` targeting (#508), `outcome: failure` on RPC throw (still audits), and access-denied (no abort, no audit). `use-ws-runtime.test.ts` (+3): the abort frame is sent (with `chatId` when scoped), and `isRunning` flips optimistically.
- **Integration E2E** — `19-chat-stop-button.spec.ts` against **real OpenClaw 2026.6.5**: send a genuinely incremental reply, stop it mid-stream, and assert (1) the full reply never lands, (2) a `chat.run_aborted` audit row appears, and (3) **the same session is immediately reusable** — a second message gets a real reply. That third assertion is the openclaw#42172 regression guard; it would go red if the upstream fix ever regressed. **Verified green locally** against the production-image integration stack (run with `DB_PASSWORD` set, as CI does).
- **Local gates**: `tsc --noEmit` clean, `eslint` 0 errors, full web unit suite (6203 passed), `test:db` (115 passed).

## Notes

- `chat.run_aborted` detail: `{ agent:{id,name}, sessionKey, runId|null, reason:"user_request" }` — no PII; sessionKey embeds the user UUID, not an email.
- A type error (`onCancel` must be `() => Promise<void>` per assistant-ui) was caught by the production-image build's stricter typecheck and fixed — folded into the feature commit.


clemenshelm:
Addressed review observation #1: `chat.run_aborted` is now emitted **only when a run was actually in flight**.

- `handleAbort` still always reaches `chatAbort` (a safe no-op on the gateway), so the stop button is never a silent no-op — even for a run the in-memory registry lost track of (e.g. after a Pinchy restart).
- But it writes the audit row only when `activeRuns.get(sessionKey)` returns a run. A stop click with nothing streaming (or a spurious/duplicate abort frame) no longer produces a `runId: null` row that a buggy or hostile client could spam.
- Audit `detail.runId` is therefore always a real runId now; docs updated accordingly.
- Unit test flipped to the new contract (no-run → `chatAbort` called, no audit). Folded into the existing commits; tsc/lint/client-router suites green.

Issue #602, opened 2026-06-26
feat(web): governed real-browser tool (JS-rendering) for agents on OpenClaw's native browser


clemenshelm:
## Why

`pinchy_web_fetch` reads **static HTML only** — it does not execute JavaScript. Two common page classes are therefore unreadable by a static fetch:

- **JavaScript single-page apps** (e.g. Canva-hosted sites like `zaruba-edv.at`): the HTML is an empty shell; content is rendered client-side.
- **Pages behind anti-bot protection** that require a real browser fingerprint / JS challenge.

PR #601 already fixed the *extraction* side (content-rich static pages like `ambersearch.de` now work, and SPA shells are reported honestly instead of dumping markup). What remains is reading pages that genuinely require a browser.

## What

OpenClaw 2026.6.8 ships a **native `browser` tool** (real Chromium via CDP/Playwright: navigate, click, type, snapshot, screenshot). We do not build a browser — we put Pinchy's governance layer on top of OpenClaw's, consistent with how we treat the runtime everywhere else.

## Scope / design questions (to resolve before implementation)

- **Mode**: agentic (act/observe/extract with LLM reasoning) vs. deterministic (navigate + read text)?
- **Sessions**: support authenticated/login sessions & cookies, or always a fresh anonymous context?
- **Downloads**: allowed, or read-only?
- **Per-agent gating**: single toggle like "Fetch web pages" + per-agent domain allow/deny (reuse the existing `WebSearchPermissionSection` wiring).
- **Domain/SSRF restriction (layered)**: in-engine request interception, an **egress proxy allowlist** (cannot be bypassed by JS), and **network-level isolation** (separate container + egress firewall blocking RFC1918 / link-local / metadata IPs). Note: the static-fetch DNS-pin SSRF defense does **not** transfer to a real browser — it does its own DNS/connections.
- **Isolation**: run the browser in a separate, network-isolated container (OpenClaw supports remote CDP: Browserless/Steel/Browserbase). Licenses: Playwright/Steel (Apache), browser-use/Stagehand (MIT) are AGPL-compatible; Browserless (SSPL) is not.
- **Audit**: a browser needs more than dispatch audit — every navigation, the domains its sub-requests touch, and downloads. Extend `pinchy-audit` accordingly.
[...]

PR #603, opened 2026-06-26
fix(web): deny OpenClaw group:ui so the native browser tool isn't silently reachable

clemenshelm:
## Problem

OpenClaw's native `browser` (and `canvas`) tools live in tool group **`group:ui`** (per `docs/gateway/config-tools.md`, OpenClaw 2026.6.8). Pinchy's per-agent deny-list `computeDeniedGroups` denied `group:runtime` / `group:fs` / `group:web` + `image_generate` — but **not `group:ui`**. So the real browser was never gated for any agent, even though Pinchy exposes no `browser`/`canvas` tool. The `build.ts` comment explicitly (and wrongly) claimed the deny-list already covered it.

**Severity: latent, not actively exploitable today** — the production image ships no Chromium, so the tool would fail at runtime. But that's an accident, not a control, and it must be a real gate before we add a governed browser feature (#602).

This was found while reviewing #601.

## Fix

- Add `group:ui` to `ALL_GROUPS` so every agent's `tools.deny` covers `browser` + `canvas`.
- **Drift guard** (`tool-registry.test.ts`): transcribes OpenClaw's built-in tool→group membership and asserts every built-in a governed agent must never reach (`exec`, `read`, `web_fetch`, `browser`, `canvas`, …) is denied — by group or standalone. This would have caught `browser` slipping through, and forces a conscious decision when a future OpenClaw upgrade adds or moves a powerful built-in.
- Correct the `build.ts` comment to describe the actual gating mechanism.
- Update the two exact `tools.deny` assertions in `openclaw-config.test.ts` to include `group:ui`.

## Tests

- TDD: drift guard written first (red — `browser`/`canvas` uncovered), then `group:ui` added (green).
- Full web unit suite: **6554 passed, 13 skipped, 0 failures**.

## Notes

Denying `group:ui` is safe: Pinchy exposes neither `browser` nor `canvas`. Prerequisite for #602 (governed real-browser feature).

Built on OpenClaw (https://docs.openclaw.ai). Pinchy: https://heypinchy.com

### DIFF
--- a/packages/web/src/__tests__/lib/openclaw-config.test.ts
+++ b/packages/web/src/__tests__/lib/openclaw-config.test.ts
@@ -671,7 +671,7 @@ describe("regenerateOpenClawConfig", () => {
       model: "anthropic/claude-opus-4-7",
       workspace: "/root/.openclaw/workspaces/uuid-agent-1",
       tools: {
-        deny: ["group:runtime", "group:fs", "group:web", "image_generate"],
+        deny: ["group:runtime", "group:fs", "group:web", "group:ui", "image_generate"],
         fs: { workspaceOnly: true },
       },
       heartbeat: { every: "0m" },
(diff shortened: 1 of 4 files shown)


clemenshelm:
Review note: an independent review confirmed this fix is correct (verified `browser`/`canvas` ∈ `group:ui` against OpenClaw 2026.6.8 docs; denying it breaks nothing — Pinchy exposes neither). It also surfaced a **broader, related gap**: Pinchy's deny-only posture leaves other powerful built-ins reachable (`cron`, `gateway`, `subagents`, `message`, media generators). That is intentionally **out of scope here** (it needs an allowlist-posture change + E2E verification, behavior risk) and is tracked in #605. This PR remains a correct, self-contained step that closes the `browser` hole.

PR #616, opened 2026-06-30
feat(pinchy-odoo): resolve nested m2o in one2many command tuples (#615)

### LINKED ISSUE #615: pinchy-odoo: scope/resolve nested m2o fields inside one2many command tuples (e.g. account.move line_ids account_id)
## Context

Follow-up to #614, which fixed top-level many2one resolution (bare \`_pinchy_ref\` + company-scoped name lookups with the OR-with-\`false\` pattern).

\`normalizeMany2OneValues\` only resolves **top-level** m2o fields of the create/write values. Nested one2many command tuples (e.g. \`account.move\` created with \`line_ids: [[0, 0, { account_id: \"…\", debit: …, credit: … }]]\`) are passed through to Odoo verbatim — the cross-company guard (\`assertNoCrossCompanyRefs\)) and the company-scoping both skip them (documented in the \`assertNoCrossCompanyRefs\` comment).

## Gap

\`account.account\` is company-scoped and can collide across companies (same account code/name in two companies). An agent creating an \`account.move\` with inline \`line_ids\` and an \`account_id\` passed as a display name (or a bare \`_pinchy_ref\`) inside the line tuple gets no scoping or ref-decoding from the plugin:

- a display name is not name-searched by Odoo on m2o write values → Odoo rejects it (expected int/[id,name]/false);
- a bare \`_pinchy_ref:…\` string is not decoded by Odoo → rejected;
- a raw numeric id works (the plugin's anti-enumeration guard only applies to top-level normalized fields, so numeric ids inside line tuples reach Odoo directly).

So the multi-company collision the top-level fix addresses does not extend one nesting level down.

## Proposal

Walk one2many command tuples (create \`[0, 0, {values}]\` and update \`[1, id, {values}]\`) in \`normalizeMany2OneValues\`, resolve their m2o fields with the same \`scopeCompanyId\` (OR-with-\`false\`), and decode bare \`_pinchy_ref\` strings. Reuse the existing \`decodeAndValidateRef\` / \`searchRelationByName\` primitives.

## Open question to resolve first

Production audit (#614) showed \`odoo_create account.move\` calls with \`line_ids: [[0, 0, { account_id: \"pinchy_ref:v1:…\", … }]]\` recorded as \`success: true\` in the audit log. Odoo should reject a \`pinchy_ref\` string for a many2one \`account_id\`. Clarify whether:

- the audit \`success\` flag reflects dispatch rather than Odoo acceptance (so the create actually failed in Odoo), or
- some path currently decodes/accepts these tokens inside line tuples.

Verify against the real mock (extend \`config/odoo-mock/server.js\` \`account.move\`/\`account.move.line\` if needed) before implementing, so the fix targets the actual behavior.

## Out of scope of #614

Tracked separately because it is a deeper change (nested command-tuple walking) and #614 stays focused on the top-level journal-resolution blocker it shipped to fix.

clemenshelm:
## Problem

PR #614 fixed **top-level** many2one resolution (bare `_pinchy_ref` + company-scoped name lookups with OR-with-`false`). Nested m2o fields inside one2many command tuples — `account.move` created with `line_ids: [[0, 0, { account_id: "<bare pinchy_ref>", debit, credit }]]` — still passed **verbatim** to Odoo. A bare ref the agent copied from `odoo_read` reached Odoo as a string where an integer id was expected → ValidationError. The nested path didn’t get the ref-decoding or company scoping the top-level fix introduced.

Tracked in #615 (with an open question about a production audit showing `success: true` for such creates — see **Audit-mystery** below).

## Fix

**Core — recursive nested resolution.** `normalizeMany2OneValues` becomes a thin wrapper delegating to `normalizeValuesRecord` + `normalizeCommandTuples`. Only Command codes `0` (create) and `1` (update) carry a values dict; `2/3/4/5/6` pass through untouched. The parent’s `scopeCompanyId` threads down (a line’s own `company_id` overrides for its siblings). Nested m2o values reuse `resolveRelationValue` unchanged — bare ref, `{ref}` object, name/code lookup, OR-with-`false` company scope all covered. Recursion capped at `MAX_NORMALIZE_DEPTH=8`.

**Cross-company guard.** `assertNoCrossCompanyRefs` now recurses into `[0,0,{values}]` / `[1,id,{values}]` tuples via `readRefCompanyTag` (already bare-string + `{ref}` capable), reusing the parent’s `company_id` as `intended` and threading a path like `line_ids[0].account_id` into the error.

**Nested-relation permissions (refined enforce).** Inline `[0,0,{…}]` line creation is part of the parent’s atomic create — already gated by the top-level `account.move:create` check — so it needs **no** separate `account.move.line:create` grant. This preserves the bookkeeper’s *“lines only inline via invoice_line_ids, never standalone”* atomicity design (`account.move.line:create` stays ungranted — test `agent-templates.test.ts:1031` keeps passing). `[1,id,{…}]` line edits require `account.move.line:write`; `[2]/[3]` line record deletion requires `account.move.line:delete` (no template grants delete — the *“never delete”* audit-trail principle; agents clear/replace lines via `[5]`/`[6]`, which rewire the relation without deleting records and need no grant). This is consistent with **every** existing Odoo template grant — **no template migration needed**. (Decision: refined enforce — see #615 discussion; literal enforce would have required granting `account.move.line:create`, opening the standalone path the AGENTS.md forbids.)

**Mock.** Adds `account.account` + `account.move.line` models, `line_ids` + `invoice_line_ids` one2many on `account.move`, `ir.model` rows, and a multi-company account collision (mirroring `account.journal`). The create handler materializes `[0,0,{values}]` tuples into `account.move.line` child records with a `move_id` back-link and **validates m2o field types** — rejecting bare-string m2o values the way Odoo does.

## Audit-mystery (#615 open question)

Production audit rows showed `tool.odoo_create` on `account.move` with `line_ids: [[0,0,{account_id: "pinchy_ref:v1:…"}]]` recorded as `success: true`. Code analysis: `client.create` throws `OdooError` on Odoo ValidationError → `errorResult` → `isError:true` + `details.error` → audit `outcome: failure`. So a create where Odoo rejects a bare-string `account_id` would be `failure`, not `success`. The mock now confirms the Odoo-faithful contract: a bare-string m2o value reaching Odoo **is** rejected (integration test `mock (Odoo-faithful) rejects a bare-string account_id`). The plugin decodes bare refs before they reach Odoo, so the success path works. The historical `success: true` rows remain unexplained by current code + Odoo behavior — likely predating the verbatim nested pass-through or a different Odoo config. I’ll update #615 with this finding; production data cleanup (moves possibly created with missing `account_id`) should be assessed against real Odoo separately.

## Tests (TDD)

- **Unit (+7)**: bare ref nested decode; nested name lookup scoped; cross-company nested rejected; non-0/1 codes untouched; inline `[0]` needs no line grant; `[1]` needs write; `[2]` needs delete; line `company_id` scope override.
- **Integration (+5)** against the real mock-odoo: bare-ref line materialized + read-back; scoped nested name; cross-company rejected end-to-end; `[1]` write-permission; mock rejects bare-string m2o (audit-mystery verification).

## Verification

- Plugin suite: **295 passed** (was 283)
- Web suite: **6593 passed** (no mock regressions)
- agent-templates: **188 passed** (no template regression — no migration needed)
- `test:scripts`: **182 passed**
- `tsc --noEmit`: clean

## Out of scope

- `odoo_read` one2many wrapping / line `_pinchy_ref` emission on create response (Odoo `create` returns only the top-level id; separate read-side enhancement).
- `many2many` command tuples (only one2many descended).
- Production data cleanup of moves created with missing `account_id` (assess after confirming the audit-mystery against real Odoo).

Closes #615.

### DIFF
--- a/config/odoo-mock/server.js
+++ b/config/odoo-mock/server.js
@@ -381,6 +381,76 @@ const MODEL_FIELDS = {
       readonly: false,
       relation: "res.partner",
     },
+    line_ids: {
+      string: "Journal Items",
+      type: "one2many",
+      required: false,
+      readonly: false,
+      relation: "account.move.line",
+    },
+    invoice_line_ids: {
+      string: "Invoice Lines",
+      type: "one2many",
+      required: false,
+      readonly: false,
+      relation: "account.move.line",
+    },
+  },
+  // Chart of accounts — company-scoped (optional company_id; false = shared).
+  // Seeded with a multi-company name collision mirroring account.journal, so
+  // nested account_id name lookups can exercise the OR-with-false scope.
+  "account.account": {
+    id: { string: "ID", type: "integer", required: false, readonly: true },
+    code: { string: "Code", type: "char", required: false, readonly: false },
+    name: { string: "Name", type: "char", required: false, readonly: false },
+    display_name: {
+      string: "Display Name",
+      type: "char",
+      required: false,
+      readonly: false,
+    },
+    account_type: {
+      string: "Type",
+      type: "char",
+      required: false,
+      readonly: false,
+    },
+    company_id: {
+      string: "Company",
+      type: "many2one",
+      required: false,
+      readonly: false,
+      relation: "res.company",
+    },
+  },
+  // Journal items (the lines of a move). account_id is the m2o whose nested
+  // resolution (#615) this model exists to test.
+  "account.move.line": {
+    id: { string: "ID", type: "integer", required: false, readonly: true },
+    name: { string: "Label", type: "char", required: false, readonly: false },
+    debit: { string: "Debit", type: "float", required: false, readonly: false },
+    credit: { string: "Credit", type: "float", required: false, readonly: false },
+    account_id: {
+      string: "Account",
+      type: "many2one",
+      required: true,
+      readonly: false,
+      relation: "account.account",
+    },
+    move_id: {
+      string: "Move",
+      type: "many2one",
+      required: true,
+      readonly: false,
+      relation: "account.move",
+    },
+    company_id: {
+      string: "Company",
+      type: "many2one",
+      required: false,
+      readonly: false,
+      relation: "res.company",
+    },
   },
 };
 
@@ -534,6 +604,8 @@ function getDefaultRecords() {
       { id: 9, model: "res.company", name: "Company" },
       { id: 10, model: "account.journal", name: "Journal" },
       { id: 11, model: "account.move", name: "Journal Entry" },
+      { id: 12, model: "account.account", name: "Account" },
+      { id: 13, model: "account.move.line", name: "Journal Item" },
     ],
     "res.users": [
       { id: 2, name: "Mitch Admin", login: "admin" },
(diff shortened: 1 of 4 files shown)


clemenshelm:
Closing as **superseded by #693**.

The #615 core (nested many2one in one2many command tuples + schema cache + faithful mock) landed on `main` via the parallel PR #627 while this PR was open — so #616 became a structurally-incompatible duplicate of an already-merged feature, which is why it now shows merge conflicts.

Rather than force-rebase this branch over `main` (which would have dropped main's `move_type` selection validation and company-scoped duplicate-invoice guard that live in the same functions), the genuinely net-new value from this PR — **nested-permission gating**, **many2many command-tuple resolution**, and **read-side one2many line refs** — has been re-implemented cleanly on top of main's structure in #693, preserving main's fixes. The review hardening done here (round-trip ref decode, guard id-position coverage, per-line company scoping, fail-loud on empty schema, mock code-1 fidelity) carried over into #693 as tested behavior.

No work lost — see #693.

PR #617, opened 2026-06-30
feat(permissions): per-agent read-only toggle (#571)

### LINKED ISSUE #571: Per-agent read-only toggle (one switch to restrict an agent to safe tools)
## Context

Today an agent becomes "read-only" only implicitly — by granting no write/powerful tools across each integration section separately (Odoo read-only dropdown, email read/draft/send, etc.). There is no single switch that forces a whole agent to read-only across every integration at once.

A simple per-agent read-only toggle is easier for customers to reason about than a per-operation permission matrix, and it's a crisp governance affordance ("this agent can look but not touch"). MuK AI has exactly this (`read_only` field, "Restrict to Read-only Tools", enforced via MCP scope) — [public source](https://github.com/muk-it/odoo-modules/tree/19.0/muk_ai).

## Current state in Pinchy

- No agent-wide toggle: the `agents` table (`schema.ts:121-143`) has only `allowedTools` as a permission-bearing column; "read-only" is an emergent property of which tools are in that array.
- Per-integration only: Odoo has a per-connection read-only/read-write/full dropdown (`odoo-permission-section.tsx:37`), email is per-operation. The permissions UI concatenates these independently — no single read-only enforcement.
- The building block exists: `getToolsByCategory("safe")` (`tool-registry.ts:239`) already returns exactly the read-only tools (we have a `category: "safe" | "powerful"` taxonomy).

## Proposal

1. Add a `readOnly` boolean to `agents` (+ migration).
2. Add a switch to the agent permissions UI.
3. In config build, when `readOnly` is true, filter the agent's tools to `safe`-category only (drop all `powerful` tools regardless of per-integration settings).

Thin layer over existing infrastructure. Fits the "auto-default + override" philosophy.

## Priority

Medium-high. Small effort, strong governance/UX value, directly on the "agent permissions & control" core differentiator — an auditable read-only guarantee is an easy sell for the sovereign/regulated positioning.

_Source: competitive code audit of MuK AI, 2026-06._

clemenshelm:
Closes #571.

## What

A single **Read-only mode** switch on the Permissions tab that restricts an agent to read-only tools. When enabled, the OpenClaw config build drops every powerful tool from the agent's allow-list — a crisp "look, don't touch" guarantee that overrides per-integration write grants at the outer allow-list boundary.

## Why

Today an agent becomes "read-only" only implicitly, by granting no write/powerful tools across each integration section separately. There is no single switch. A per-agent read-only toggle is easier for customers to reason about than a per-operation permission matrix, and it's a crisp governance affordance directly on Pinchy's core differentiator (agent permissions & control). An auditable read-only guarantee is an easy sell for the sovereign/regulated positioning.

## How (enforcement)

The build layer is the real guarantee, not the UI: when `agent.readOnly` is set, the new `filterReadOnlyToolIds` helper filters the per-agent `tools.allow` list to safe-category tools only. Because OpenClaw treats `allow` (with no profile) as an absolute allow-list (`full ∩ allow`), every powerful tool is denied regardless of what the per-integration plugin config grants — Odoo create/write/delete, email draft/send, file writes, and web search/fetch all drop out. Read-only tools (file list/read, Odoo read/count/aggregate/list/describe, email list/read/search) and the read-only built-ins (memory, PDF/image reading, session status) survive.

The same filter is applied to the downstream build conditionals (write_paths for `pinchy_write`, context tools, web plugin config) so a read-only flag flip can't leave stale write wiring in place.

## Changes

- `agents.read_only` boolean column + Drizzle migration `0044_tan_warstar` (defaults `false`; recreates the `active_agents` view to include the column).
- `filterReadOnlyToolIds` in `tool-registry.ts` — drops tools categorized as `powerful`, keeps safe tools and unclassified tools (built-ins are read-only by construction).
- `build.ts` applies the filter to the per-agent allow-list and to the web-plugin-config detection.
- `PATCH /api/agents/[agentId]` accepts `readOnly`, gated admin-only and blocked for personal agents (same rule as `allowedTools`), with an audit diff entry and OpenClaw config regen.
- Permissions UI: a Read-only mode switch at the top of the Permissions tab; powerful KB/web checkboxes and the Odoo/Email sections are disabled while it's on, so the UI reflects what the agent can actually do. The emitted `allowedTools` is filtered to safe-only so the stored value matches what the build emits.
- Docs: documents read-only mode in `concepts/agent-permissions.mdx`.

## Tests

- `filterReadOnlyToolIds` unit tests (drops powerful, keeps safe + unclassified, full-allowlist read-only guarantee).
- Permissions component: powerful checkbox disabled + powerful tools dropped from emitted `allowedTools` when readOnly is on; toggle renders for admins.
- API route: admin can enable, non-admin gets 403, personal agent gets 400.
- Existing permissions/page/API tests updated for the new `readOnly` field in `PermissionsValues` / the agent shape.

## Notes / out of scope

- `pinchy_save_*` context tools are not in `TOOL_REGISTRY` (no category), so they survive the read-only filter. They save user context/memory rather than touching external systems, so this matches the issue's literal "safe-category only" proposal; classifying them is a separate decision.
- The build-time filter is the guarantee; the UI disabling is for clarity. The per-integration plugin configs still carry their granted operations, but the allow-list denies the powerful tools at the runtime boundary.

### DIFF
--- a/docs/src/content/docs/concepts/agent-permissions.mdx
+++ b/docs/src/content/docs/concepts/agent-permissions.mdx
@@ -74,6 +74,14 @@ The Permissions tab shows all available tools grouped by category. Check or unch
 The Permissions tab is only visible to admins and only for shared agents. Personal agents (the auto-created Smithers each user gets) don't have this tab — they run with default tools only.
 :::
 
+### Read-only mode
+
+At the top of the Permissions tab there is a **Read-only mode** switch. Turning it on restricts the agent to read-only tools in one step — a crisp "this agent can look, but not touch" guarantee that is easy to reason about and to audit.
+
+When read-only mode is on, every powerful tool is dropped from the agent's OpenClaw allow-list at config-build time, regardless of what the per-integration sections below grant: file writes (`pinchy_write`), web search and fetch, Odoo create/write/delete and the governed action tools, and email draft/send. The read-only tools (file listing/reading, Odoo read/count/aggregate/list/describe, email list/read/search) and the read-only built-ins (memory, PDF/image reading, session status) stay available.
+
+This is a layer on top of the per-tool checkboxes, not a replacement for them. It overrides them: even if an admin has granted `pinchy_write` or Odoo write access in the sections below, a read-only agent cannot use those tools. Turn the switch off to configure write access again. The per-integration sections are disabled while read-only mode is on so the UI reflects what the agent can actually do.
+
 For the full list of tabs in Agent Settings and who sees which, see [Agent Settings](/concepts/agent-settings/).
 
 ### Configuring directory access
--- a/packages/web/drizzle/0044_tan_warstar.sql
+++ b/packages/web/drizzle/0044_tan_warstar.sql
@@ -1,0 +1,3 @@
+DROP VIEW "public"."active_agents";--> statement-breakpoint
+ALTER TABLE "agents" ADD COLUMN "read_only" boolean DEFAULT false NOT NULL;--> statement-breakpoint
+CREATE VIEW "public"."active_agents" AS (select "id", "name", "model", "template_id", "plugin_config", "allowed_tools", "skills", "owner_id", "is_personal", "read_only", "visibility", "greeting_message", "tagline", "avatar_seed", "personality_preset_id", "created_at", "deleted_at" from "agents" where "agents"."deleted_at" is null);
--- a/packages/web/drizzle/meta/0044_snapshot.json
+++ b/packages/web/drizzle/meta/0044_snapshot.json
@@ -1,0 +1,2171 @@
+{
+  "id": "c3838292-587c-46ea-883f-12dc4fb46b5a",
+  "prevId": "81ef12b1-d31e-4c66-90e7-15aee993bac8",
+  "version": "7",
+  "dialect": "postgresql",
+  "tables": {
+    "public.account": {
+      "name": "account",
+      "schema": "",
+      "columns": {
+        "id": {
+          "name": "id",
+          "type": "text",
+          "primaryKey": true,
+          "notNull": true
+        },
+        "user_id": {
+          "name": "user_id",
+          "type": "text",
+          "primaryKey": false,
+          "notNull": true
+        },
+        "account_id": {
+          "name": "account_id",
+          "type": "text",
+          "primaryKey": false,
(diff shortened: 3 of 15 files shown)


clemenshelm:
Closing after a market/governance research pass on whether a **global cross-integration per-agent read-only master toggle** is a real customer need. Conclusion: it is closer to feature-envy than a genuine governance requirement, so we're not shipping it in this form.

**Why**

1. **Not a market standard.** No comparable platform ships a cross-integration read-only master switch. Copilot Studio and Odoo 19 gate writes per-tool/per-connector via RBAC + data policies; agents can still write. The only real precedent, the GitHub MCP Server `--read-only` flag, is a *single-vendor, single-integration* server flag — i.e. exactly what our per-integration Odoo read-only dropdown already is, not a cross-integration master toggle.

2. **Governance frameworks want the opposite of a coarse toggle.** EU AI Act Art. 14, NIST AI RMF and ISO 42001 call for risk-calibrated, granular per-action control + human-in-the-loop for write/irreversible actions + auditability. A binary flag satisfies none of these; our existing per-integration / per-operation scopes + tamper-evident audit already do. Practitioner guidance (Okta, Zenity, Noma, Truto, WorkOS) explicitly calls coarse all-or-nothing gating an anti-pattern.

3. **It contradicts our core use case.** "Read Odoo and *create quotes*" is a write operation. A read-only agent can't create quotes, so the toggle is unusable for the primary use case and only fits dedicated pure reporting/analysis agents — a narrow niche.

4. **The enforcement here is fail-open.** `filterReadOnlyToolIds` is a denylist: anything not explicitly categorized `powerful` survives. `pinchy_save_org_context` / `pinchy_save_user_context` write and are not in `TOOL_REGISTRY`, so a "read-only" agent still writes context — and there's no drift guard forcing future write-tools to be classified. An "auditable read-only guarantee" with holes is worse than none in a CISO pitch.

**What we do instead:** prioritize human-in-the-loop approval for risky writes (#124 / PR #517). That is the control model frameworks require and CISOs actually buy, and it's already half-built.

The requirement originated from the MuK AI code audit (#571), not from a customer. If a dedicated read-only *reporting-agent preset* is ever wanted, it should be a fresh, fail-closed, narrowly-positioned feature — not this cross-integration guarantee.

Issue #338, opened 2026-05-11
Gemini 3 tool calls fail intermittently with 'missing thought_signature' (upstream OpenClaw bug)


clemenshelm:
## Summary

Agents configured with `ollama-cloud/gemini-3-flash-preview` (and presumably other Gemini 3 thinking models) intermittently fail with a generic-looking error in the chat UI:

> **Piper couldn't respond**
> LLM request failed: provider rejected the request schema or tool payload.

The user can hit **Retry** and the same message often succeeds. So the impact is annoyance, not data-loss — but the wording is misleading (it sounds like *we* sent a malformed request, when in fact the upstream OpenClaw → Gemini path is dropping the `thought_signature` field).

## Root cause (upstream)

Tracked at OpenClaw [openclaw/openclaw#72879](https://github.com/openclaw/openclaw/issues/72879). Gemini 3 requires every `functionCall` part on subsequent turns to carry the `thought_signature` it returned earlier; OpenClaw 2026.4.27's Gemini provider adapter (specifically the OpenAI-compatible path used by `ollama-cloud`) drops it on some replay paths. Comment with our 4.27 reproduction posted at [openclaw#72879 (comment)](https://github.com/openclaw/openclaw/issues/72879#issuecomment-4423525729).

## Evidence from production

Two independent occurrences today, two different agents, same model:

| Time (UTC) | Ref ID | Notes |
|---|---|---|
| 2026-05-11 15:41 | `3d5cf450-…` | text-only conversation |
| 2026-05-11 16:03 | `8f8ceafd-…` | message with image + PDF attachment |

The earlier hypothesis that this was an image-plus-PDF interaction bug was wrong — the image+PDF case just happened to be the second occurrence. Both are pure Gemini-side schema rejections on the tool-call round-trip.

## Why this matters for our users

- Error wording (\"the request schema or tool payload\") implies our code is at fault. It isn't.
- Modal users have no way to know that retry is a reliable workaround.
- Gemini 3 Flash is auto-pickable as the default model for new agents in some configurations (see [resolveDefaultPdfModel](packages/web/src/lib/openclaw-config/build.ts) and the agent template defaults), so new users could land on this without ever picking the model themselves.

## Decision deferred

We discussed but **deliberately did not** add an auto-retry in Pinchy yet:

- Vercel AI SDK, Anthropic SDK, LangChain all explicitly avoid auto-retry on 400/schema errors because it masks real bugs.
[...]

PR #621, opened 2026-06-30
fix(chat): rewrite generic provider-rejection envelope + dedicated audit class (#584)

### LINKED ISSUE #584: Chat error banner mislabels billing/credit failures as 'schema or tool payload' rejection
## Symptom
When an LLM provider rejects a run due to **billing/credit exhaustion** (e.g. Anthropic: "Your credit balance is too low"), the durable paused-error banner shows the misleading generic message **"LLM request failed: provider rejected the request schema or tool payload."** and Pinchy stores `error_class: "unknown"`. This contradicts the feature's promise of *honest, cause-specific wording*.

## Root cause
OpenClaw surfaces the generic *"provider rejected the request schema or tool payload"* string in the **error chunk** Pinchy consumes (`client-router.ts:~1154`, `chunk.text`), while the billing cause (`reason=billing`, *"credit balance too low"*) appears only in OpenClaw's internal failover-decision logs and the WS-level `FailoverError` errorMessage. So `classifyAgentError` (`packages/web/src/server/agent-error-classifier.ts`) never sees the billing wording — even though `PROVIDER_CONFIG_PATTERN` (`error-patterns.ts`) already matches `credit|balance|quota`.

## Proposed fix
Classify from the WS-level `FailoverError` errorMessage (which contains "credit balance too low") in addition to `chunk.text`, OR map OpenClaw's `errorCode`/billing signal to the existing `provider_config` class so the banner shows the right cause + the "check Settings > Providers / billing" hint. Partly upstream (OpenClaw wording).

Non-blocker; fast-follow after v0.7.0.

clemenshelm:
Closes #584.

## What

When an LLM provider rejects a run for an **account-side reason** (depleted credit, invalid key, exhausted quota), OpenClaw collapses the cause and the error chunk Pinchy receives carries only the generic `"LLM request failed: provider rejected the request schema or tool payload."` wording — the real cause never reaches Pinchy in `chunk.text`. Two problems resulted:

1. The durable paused-error **banner** showed that raw wording, which reads like a malformed-request bug.
2. The audit row bucketed it as `error_class: "unknown"`, indistinguishable from truly unrecognised strings.

## Fix

- **Banner text:** `presentProviderError` now rewrites the bare generic envelope to an honest, actionable message: _"The AI provider rejected the request. This is often a provider-account issue (billing, API key, or quota) — check Settings > Providers."_ Applied at both the live error frame and the durable-banner route (`active-error/route.ts`), which already call `presentProviderError`. The audit trail keeps the raw text.
- **Audit class:** `classifyAgentError` gains a dedicated `provider_rejected_generic` class. It is honest — not `provider_config` (which would assert an unproven cause in the append-only, HMAC-signed audit trail) and not `unknown` (which buckets it with unrecognised strings). Operators get a distinguishable bucket for the generic collapse.
- **Precedence preserved:** cause-specific wording (credit/balance) still classifies as `provider_config`; a `thought_signature` schema rejection still classifies as `schema_rejection`. The generic check runs after `provider_config` so a chunk carrying both the envelope AND cause-specific wording still classifies by the cause.
- **Guard:** `presentProviderError` only rewrites the generic envelope when it is NOT carrying a `thought_signature` (the Gemini-3 schema rejection carries the same envelope text plus `thought_signature`, and has its own user-facing handling via `upstreamFormatError`). Without this guard the schema-rejection wording would be collapsed into the account-issue message.
- Shared `thought_signature` patterns moved to `error-patterns.ts` (`isThoughtSignatureRejection`) so the classifier and the rewriter agree, preserving the original narrow anchoring (snake_case with `i`, camelCase requiring the capital `S` so a bare `thoughtsignature` can't hijack the branch).

## Out of scope (upstream-gated)

Showing the **specific** billing cause (e.g. "credit balance too low") in the banner requires openclaw-node to expose the FailoverError `errorMessage`/`errorCode` on the `ChatChunk` error — today the chunk only carries `text`, which is the collapsed generic wording. That's an openclaw-node change; this PR does the honest Pinchy-side improvement without asserting a cause it can't see.

## Tests

- `agent-error-classifier.test.ts`: generic envelope → `provider_rejected_generic`; thought_signature payload still → `schema_rejection`; credit/balance still → `provider_config`.
- `error-hints.test.ts`: `presentProviderError` rewrites the bare generic envelope (no "schema or tool payload" reaches the user; message names the account-cause family); the thought_signature payload passes through unchanged.
- Full unit suite green; `tsc --noEmit` clean.

### DIFF
--- a/packages/web/src/__tests__/server/agent-error-classifier.test.ts
+++ b/packages/web/src/__tests__/server/agent-error-classifier.test.ts
@@ -88,6 +88,34 @@ describe("classifyAgentError", () => {
     expect(classifyAgentError("")).toBe("unknown");
   });
 
+  it("classifies OpenClaw's generic provider-rejection envelope as provider_rejected_generic (#584)", () => {
+    // When a provider rejects a run for an account-side reason OpenClaw
+    // collapses (e.g. depleted credit), the error chunk carries exactly this
+    // wording and nothing cause-specific. The real cause is unknowable from
+    // this text, so it gets its own honest audit class — not `unknown` (which
+    // buckets it with truly unrecognised strings) and not `provider_config`
+    // (which would assert an unproven cause in the append-only audit trail).
+    const text = "LLM request failed: provider rejected the request schema or tool payload.";
+    expect(classifyAgentError(text)).toBe("provider_rejected_generic");
+  });
+
+  it("still classifies the thought_signature payload as schema_rejection, not provider_rejected_generic (#584)", () => {
+    // The Gemini-3 schema-rejection payload carries the generic envelope text
+    // AND a thought_signature — the specific schema signal must win so the
+    // audit class stays schema_rejection.
+    const text =
+      "LLM request failed: provider rejected the request schema or tool payload. " +
+      'rawError=400 "Function call is missing a thought_signature in functionCall parts."';
+    expect(classifyAgentError(text)).toBe("schema_rejection");
+  });
+
+  it("still classifies a credit/balance wording as provider_config, not provider_rejected_generic (#584)", () => {
+    // If the cause-specific wording DOES reach the chunk (credit/balance), it
+    // must classify as provider_config — the generic class is only for the
+    // cause-unknowable envelope.
+    expect(classifyAgentError("Your credit balance is too low")).toBe("provider_config");
+  });
+
   it("classifies transient before provider_config when text mentions 'exceeded'", () => {
     // Defensive: `rate limit exceeded` contains "exceeded" which would also
     // match the provider-config family. Mirrors the same precedence rule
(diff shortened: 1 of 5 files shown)


clemenshelm:
Closing as superseded by #675.

This PR was opened on 2026-06-30 and never rebased. Since then, main took a **different, deliberate** direction for the same problem: `6a59553df` reworked `presentProviderError` to take a `modelName` param and name the dispatched model instead of hand-crafting a replacement message. That change conflicts directly with this PR's rewrite and its `presentProviderError(errorText)` signature.

#675 reconciles this PR's intent onto current main:
- keeps the honest banner rewrite for the generic `"…schema or tool payload"` envelope (which main still shows verbatim, contradicting the hint), composed with the newer model-naming behavior;
- keeps the dedicated `provider_rejected_generic` audit class;
- keeps the shared `isThoughtSignatureRejection` guard so the #338 schema rejection is untouched.

Ground truth re-verified while rebasing: openclaw-node forwards `data.error` as `chunk.text` (or bare `"LLM request failed."` when empty) and still carries no `errorCode`/`errorMessage` on the ChatChunk error — so the "show the specific billing cause" part remains genuinely upstream-gated, exactly as this PR's out-of-scope note said.

Issue #477, opened 2026-06-11
feat(telegram): conflict validation + auto-disable for multi-instance bot polling (target v0.6.0)


clemenshelm:
**Not a v0.5.7 blocker. Target: v0.6.0.** Follow-up to #476.

## Context
The v0.5.7 channel-health watchdog surfaced a real cross-instance `getUpdates` conflict (staging and production both polling `@pinchy_smithers_bot` — see #476). #476 covers the removal/reload gaps; this issue is the design for **preventing** conflicts where possible and **handling** them gracefully where not.

## Research summary (what is actually possible)
- **Telegram has no lock/coordination API.** The 409 `"Conflict: terminated by other getUpdates request"` is the only signal, and the **latest** poller wins (terminates the previous long-poll). Two instances genuinely fight; neither holds a stable lock.
- **Connect-time detection is possible but racy:** a `getUpdates(timeout=1)` probe returns 409 iff another consumer is polling *right now*. Catches the common case; cannot prevent a conflict that starts later.
- **Webhooks would eliminate the problem** (exclusive single URL; multi-consumer constraint applies only to polling) — but require public HTTPS, so they cannot be the default for an offline-first, self-hosted product. Possible future opt-in for reachable deployments.
- **Prior art** (grammY, Telegraf, python-telegram-bot, n8n, Botpress, Home Assistant): none auto-handle this. Universal guidance is operational: **one bot token per environment**. Distributed locks/heartbeats don't fit offline-first (no reliable shared substrate between independent deployments).
- **Within one instance** Pinchy already rejects reusing a token across agents (`409 "This bot token is already in use by another agent"` in the connect route) — so single-instance double-polling cannot happen today.
- **OpenClaw has no granular stop-channel API** — stopping a poller means removing the account from `channels.telegram` config *and* ensuring OpenClaw actually reloads. **The #476 reload fix is therefore a prerequisite for auto-disable.**

### 1. Connect-time conflict probe (cheap, high value)
In `POST /api/agents/[agentId]/channels/telegram`, after the existing `getMe` validation + duplicate-token check: call `getUpdates(timeout=1)`. On 409, reject with an actionable inline error:
> "This bot is already being polled by another deployment (e.g. staging/production). Use a separate bot per environment, or disconnect it there first."

~15 lines next to the existing checks. Would have caught the staging incident at connect time. Documented as racy (point-in-time check), which is fine — layer 2 is the net.

### 2. Auto-disable on sustained runtime conflict (robust net)
[...]

Issue #476, opened 2026-06-10
Telegram channel management gaps surfaced by channel-health watchdog (target v0.6.0)


clemenshelm:
**Not a release blocker for v0.5.7. Target: v0.6.0.**

## How this surfaced
The new channel-health watchdog (v0.5.7) started logging a persistent flap on a staging Smithers Telegram bot — repeating `channel.degraded` → `channel.polling_failed` → `channel.recovered` every ~1–2 min, all with:
```
lastError: "Conflict: terminated by other getUpdates request;
            make sure that only one bot instance is running |
            Telegram ingress worker exited with code 1"
```
Investigating it (a) validated the watchdog — it caught a real, previously-silent problem — and (b) turned up several telegram-channel-management gaps worth fixing.

## Root cause of the surfaced conflict (operational, not a code bug)
`@pinchy_smithers_bot` is connected on **two Pinchy instances at once**:
- staging — agent `7af83d1c` (a leftover personal Smithers, `channel.created` 2026-05-04, never removed)
- production `pinchy.heypinchy.com` — agent `2f1bc2e2`

Telegram allows exactly one `getUpdates` consumer per bot token, so both instances' OpenClaw pollers terminate each other → **both flap with 409** (confirmed: production's openclaw log shows the identical restart loop). Staging is therefore also degrading the *production* Smithers bot. This is a staging-hygiene violation (staging is holding a production bot token) — the operational fix is to disconnect that bot on staging and use a throwaway bot for staging tests.

Note: Pinchy already prevents this **within one instance** — the connect route validates the token via Telegram and rejects a second agent reusing the same bot (`409 "This bot token is already in use by another agent"`, `agents/[agentId]/channels/telegram/route.ts:77-91`). There is no cross-instance protection (inherently hard; the watchdog is the right mitigation). A "2 users / 1 bot on the same instance" cause was ruled out.

### 1. Telegram removal does not promptly stop OpenClaw's poller (main one)
[...]

PR #628, opened 2026-06-30
feat(telegram): per-agent disconnect + channel_links cleanup (#476 gaps 2+3)

clemenshelm:
Part of #476 — gaps 2 and 3. (Gap 1 deferred; see below.)

## Gap 2 — per-personal-agent disconnect
A personal agent's owner can now disconnect its own Telegram bot without an admin. Previously the only path was the org-wide **"Remove Telegram for everyone"** (admin-only, removes every user's bot) — too blunt for a single user's Smithers.

`DELETE /api/agents/[id]/channels/telegram` now accepts **admins (any agent) OR the personal agent's owner**; the connect path stays admin-only. Switched from `requireAdmin` to `withAuth` so the owner-or-admin check can run (satisfies the `no-direct-session` lint rule via the wrapper). The agent-settings Telegram tab now shows the **Disconnect** button for personal agents too (was info-only text pointing at the org-wide action).

## Gap 3 — channel_links cleanup on last-bot removal
When the last Telegram bot is removed via the per-agent path, clear the user↔telegram-user `channel_links` so a later re-registered bot can't re-grant access to stale pairings without re-pairing. The org-wide removal already cleared `channel_links`; this closes the per-agent-removal path where every bot is gone individually but the links persisted. (If other bots remain, `channel_links` are left intact — they're per-user and shared across bots.)

## Deferred
**Gap 1** (a config change must promptly stop the OpenClaw poller, rather than waiting for OC's eventual inotify/restart) needs a `config.apply`/reload mechanism and is also a prerequisite for #477's auto-disable. Tracked separately to keep this PR focused.

## Tests
`agent-telegram-channel.test.ts` DELETE suite rewritten for the new auth:
- admin can disconnect a personal agent; a personal agent's owner (non-admin) can disconnect their own; a non-owner non-admin gets 403; a non-admin on a shared agent gets 403.
- gap 3: `channel_links` cleared when the last bot is removed; left intact when other bots remain.
- 404 / 401 unchanged.
Unit suite green; `tsc --noEmit` clean.

### DIFF
--- a/packages/web/src/__tests__/api/agent-telegram-channel.integration.test.ts
+++ b/packages/web/src/__tests__/api/agent-telegram-channel.integration.test.ts
@@ -1,0 +1,220 @@
+// Real-DB integration test for DELETE /api/agents/[agentId]/channels/telegram.
+//
+// The unit test in agent-telegram-channel.test.ts mocks `@/db` entirely, so it
+// only proves "db.delete was called" — not that the right rows actually
+// disappear. This test hits a real PostgreSQL (provisioned by global-setup.ts,
+// truncated between tests by setup.ts) to verify the Gap 3 (#476) hardening
+// for real:
+//
+//   - When the last Telegram bot in the instance is disconnected, ALL
+//     channel_links rows with channel = "telegram" are deleted — across every
+//     user, not just the agent being disconnected — and the existing
+//     channel.deleted audit row records how many links were cleared via
+//     detail.channelLinksCleared.
+//   - When another agent's Telegram bot remains configured, channel_links are
+//     left untouched and channelLinksCleared is 0.
+//
+// What stays mocked, and why:
+//   - @/lib/auth (getSession) / next/headers — no real browser session exists
+//     in a test process; withAuth reads the session via headers().
+//   - @/lib/openclaw-config (updateTelegramChannelConfig) — writes OpenClaw's
+//     on-disk config and flips restart-state; irrelevant to the DB invariant
+//     under test and already covered by openclaw-config's own tests.
+//   - @/lib/telegram-allow-store (clearAllowStoreForAccount,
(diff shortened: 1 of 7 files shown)


clemenshelm:
## Hardening pass (rebased on main)

Rebased on `main` (conflict-free) and addressed two review gaps before merge:

**Real-DB coverage for gap 3.** The gap-3 unit test mocked `@/db` entirely and only asserted `db.delete` was *called* — a false-green that never proved the right rows disappear. Added `agent-telegram-channel.integration.test.ts` (real Postgres, no `@/db` mock): it seeds a personal agent + bot token + `channel_links` across **multiple distinct users**, disconnects as the owner, and asserts against real `SELECT`s that every telegram `channel_link` is gone (last-bot case) or untouched (other-bot-remains case). This is the read-side invariant that matters: *last bot removed ⟹ no stale pairings a re-registered bot could auto-grant on.*

**The org-wide sweep is now attributable.** A normal member disconnecting their own personal Smithers can wipe *every* user's telegram `channel_links` if it was the last bot — previously invisible in the audit trail. `DELETE` now uses `.returning()` to count the cleared links and records `channelLinksCleared` on the existing `channel.deleted` audit row (`0` when other bots remain). No new event type — `DeleteDetail` already allows extra keys.

Gates green: unit 28/28, integration 2/2, full `test:db` 142/142, `tsc --noEmit` clean, eslint/prettier clean.

**Gap 1 stays deferred, by design** — the poller-stop mechanism (`config.apply`/reload so a removed bot stops polling promptly, rather than waiting on OC's inotify) is a distinct, higher-risk subsystem and a prerequisite for #477. It's being scoped as its own focused PR with real E2E verification against the OC stack, rather than dragging gaps 2+3 behind it.

PR #662, opened 2026-07-06
fix(chat): adopt refocus history when a leading greeting hides the landed reply

clemenshelm:
## Root cause

`shouldReplaceLocalWithServerHistory` in `packages/web/src/hooks/use-ws-runtime.ts` gates whole-list history reconcile on tab-refocus. Its #310 guard only adopted server history when it was strictly longer than the local list. A fresh session's local list carries the client-only agent GREETING (an assistant message at index 0) that the server omits from history once real turns exist. After a tab-refocus during the first turn — local ends in the acked user message (padded by the greeting to the same length as the server's `[user, assistant]` history) — `historyMessages.length > prev.length` was `2 > 2 = false`, so the completed assistant reply was silently dropped and the agent looked unresponsive.

## Fix

Purely additive: also adopt the server history when it ends in an assistant turn (mirroring the greeting-offset handling already used by the sibling function `preserveRicherLocalOverOversizedHistory`), while keeping the existing `status === "sent"` guard that protects a queued ("sending") message from being wiped.

```diff
   // lastNonError.role === "user"
-  return lastNonError.status === "sent" && historyMessages.length > prev.length;
+  const serverEndsInAssistant = historyMessages[historyMessages.length - 1]?.role === "assistant";
+  return (
+    lastNonError.status === "sent" &&
+    (historyMessages.length > prev.length || serverEndsInAssistant)
+  );
```

## Tests

- `packages/web/src/__tests__/hooks/should-replace-local-with-server-history.test.ts` — new unit test: "adopts server history when a leading greeting makes it look equal-length but the reply landed". RED before the fix (`expected false to be true`), GREEN after (13/13 passing).
- `packages/web/src/__tests__/hooks/use-ws-runtime-refocus-shrink.test.tsx` — new hook-level end-to-end test mirroring the existing renderHook + mock-WebSocket harness: fresh session with only the greeting, user sends, server acks (WS drops before the first chunk — the #310 precondition), tab backgrounds/reconnects, recovery history (no `activeRun`) holds `[user, assistant-reply]`. Asserts the reply text renders after refocus. Confirmed RED before the fix, GREEN after, with `shouldReplaceWithHistory: true` and `shouldStageReplace: true` routing through the crash-safe `stageDestructiveHistoryReconcile` path.

Found on staging 2026-07-05; blocks v0.8.0.

### DIFF
--- a/packages/web/src/__tests__/hooks/should-replace-local-with-server-history.test.ts
+++ b/packages/web/src/__tests__/hooks/should-replace-local-with-server-history.test.ts
@@ -142,4 +142,21 @@ describe("shouldReplaceLocalWithServerHistory", () => {
       false
     );
   });
+
+  // Greeting-offset regression (staging 2026-07-05): a fresh session's local
+  // list carries the client-only agent greeting (assistant at index 0) that the
+  // server omits from history once real turns exist. After a tab-refocus during
+  // the first turn — local ends in the acked user message, server history is
+  // [user, assistant] — the greeting pads prev.length to EQUAL the server
+  // history length, so the #310 strictly-longer guard alone wrongly drops the
+  // completed reply and the agent looks silent though it answered.
+  it("adopts server history when a leading greeting makes it look equal-length but the reply landed", () => {
+    expect(
+      shouldReplaceLocalWithServerHistory(
+        [assistant("Good day. How may I help?"), user("what time is it?", "sent")],
+        [user("what time is it?"), assistant("It's 07:12 UTC.")],
+        true
+      )
+    ).toBe(true);
+  });
 });
--- a/packages/web/src/__tests__/hooks/should-replace-local-with-server-history.test.ts
+++ b/packages/web/src/__tests__/hooks/should-replace-local-with-server-history.test.ts
@@ -159,4 +159,32 @@ describe("shouldReplaceLocalWithServerHistory", () => {
       )
     ).toBe(true);
   });
+
+  // Data-loss guard (review of the greeting-offset fix): adopting on a
+  // greeting offset must NEVER fire for a SHORTER server history — that would
+  // erase a just-sent local user turn the server hasn't persisted yet. The
+  // trigger is a reload during a PENDING run: the user sends a follow-up in an
+  // EXISTING chat, OpenClaw hasn't streamed the first chunk, so client-router
+  // withholds the activeRun signal (its `firstChunkAt !== null` gate). The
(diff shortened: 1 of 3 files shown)


clemenshelm:
## Follow-up commit `0a75d6e`: greeting offset instead of "ends in assistant"

A review of this branch found that the `serverEndsInAssistant` clause opens a **data-loss** path, so I replaced it with a tighter greeting-offset comparison.

**The hole:** `serverEndsInAssistant` adopts the server history for *any* frame ending in an assistant turn, regardless of length. That also fires on the **pending-follow-up reload window**: the user sends a second message in an existing chat, OpenClaw hasn't streamed the first chunk yet, so `client-router` withholds the `activeRun` signal (its `firstChunkAt !== null` gate). The recovery frame is then the *previous* turn's `[user, assistant]` — ending in an assistant — with **no** `activeRun`, and **shorter** than the rich local list. With no `activeRun` it routes through `stageDestructiveHistoryReconcile`, which (unlike the synchronous path at the `next.length < prev.length` guard) has **no** anti-shrink guard → the just-sent follow-up is silently dropped.

**The fix:** discount the client-only leading greeting from the length comparison instead of widening the gate. A persisted OpenClaw history *always* begins with a user turn (system turns are filtered out in `fetchAndParseHistory`; users open every conversation), so a leading assistant in the *local* list can only be the greeting — subtract it. This surfaces the landed reply for the greeting case **without** adopting a shorter history. Mirrors the tail-anchored greeting handling in `preserveRicherLocalOverOversizedHistory`.

```diff
-  const serverEndsInAssistant = historyMessages[historyMessages.length - 1]?.role === "assistant";
-  return (
-    lastNonError.status === "sent" &&
-    (historyMessages.length > prev.length || serverEndsInAssistant)
-  );
+  const leadingGreetingOffset = prev[0]?.role === "assistant" ? 1 : 0;
+  return (
+    lastNonError.status === "sent" && historyMessages.length > prev.length - leadingGreetingOffset
+  );
```

**Tests (both RED on the old rule → GREEN now):**
- unit: keeps local when server history is shorter than local minus the greeting (pending follow-up must not be dropped)
- hook-level: no-`activeRun` shorter recovery history must not drop a just-sent pending follow-up

The original greeting-surface tests stay green (the offset yields `true` for the equal-length greeting case). Full web unit suite: 7222 passed / 13 skipped, tsc clean, ESLint 0 errors, Prettier clean.

PR #664, opened 2026-07-06
fix(integrations): OAuth connect failures as a persistent banner, not a toast

clemenshelm:
## Problem (found on staging 2026-07-06)

An admin re-set-up the Microsoft OAuth app, logged in successfully, but got **only the app config and no integration**. Debugging the callback via SSH + the audit log pinned it to a real `token_exchange_failed` (wrong Client Secret) — Pinchy behaved correctly (deleted the pending row, audited the failure, redirected with `?error=`). But the only user-facing signal was a **transient `toast.error`** that auto-expires in ~4s — gone before a user whose attention is still on the provider's redirect can read it. So the actionable error ("double-check the Client Secret") vanished and the failure looked like a silent bug.

## Rule (why toast was wrong here)

AGENTS.md said "toast for actions that navigate away." The OAuth connect navigates away, so it formally fell into "toast" — but the rule had no category for a **permanent, actionable** error that lands *after* a redirect. That's the gap.

Best practice (LogRocket, RingCentral UX) and our own existing patterns (`cannotDecrypt`, `auth_failed`, "App not configured" are all persistent, but all require a surviving connection row — which a connect failure deletes) agree: a permanent, actionable error → **persistent inline banner**, not a transient toast.

## Change

- **`settings-integrations.tsx`**: render the `?error=` message as a **persistent, dismissible inline banner** at the top of the Integrations section (styled like the existing `cannotDecrypt` destructive alert), seeded from the prop into local state so stripping the URL doesn't clear it. Removed the `toast.error`. No toast for the same action (AGENTS.md: don't mix inline + toast).
- **AGENTS.md** "Error And Notification UI": clarified — "navigated away" alone does **not** justify a toast; classify by transient-and-retryable (toast) vs permanent-and-actionable (persistent inline banner).

## Tests (TDD, RED → GREEN)

- 7 existing OAuth-callback-error tests **converted in place** from toast assertions to banner assertions; **+2 net** (dismiss + no-error cases). All RED before the change, GREEN after.
- Full web unit suite: **7224 passed / 13 skipped** (no new skips), tsc clean, ESLint 0 errors, Prettier clean.

Not a release blocker by itself (the original failure was an Azure config issue), but it's the reason the config error was invisible — solid fix for the v0.8.0 email/OAuth surface.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -193,7 +193,9 @@ This makes contract drift between client payload and server schema a compile-tim
 
 Use inline form errors when the error is tied to a field, the user can correct the input, and the form/dialog stays open.
 
-Use toast notifications for completed actions, background/system errors, and actions that navigate away or close the dialog.
+Use toast notifications for completed actions, background/system errors, and transient errors the user can simply retry.
+
+Use a persistent, dismissible inline banner (not an auto-expiring toast) for a permanent, actionable error that lands after a full-page redirect — e.g. an OAuth connect failure surfaced via a `?error=` query param. "The action navigated away" does NOT by itself justify a toast: after a redirect the user's attention is on the provider, so a few-second toast is gone before they read it, and the error needs a configuration fix that outlives a toast. Classify by whether the error is transient-and-retryable (toast) or permanent-and-actionable (persistent inline banner), not by whether the flow navigated.
 
 Do not mix inline errors and toast errors for the same action. Success confirmations should be toasts unless a multi-step flow intentionally shows a success screen.
 
--- a/packages/web/src/__tests__/components/settings-integrations.test.tsx
+++ b/packages/web/src/__tests__/components/settings-integrations.test.tsx
@@ -737,83 +737,102 @@ describe("SettingsIntegrations — derived 'app not configured' state", () => {
   });
 });
 
-describe("SettingsIntegrations — OAuth callback errors", () => {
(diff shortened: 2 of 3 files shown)


clemenshelm:
Follow-up: addressed the two findings from the `/review 664` pass (commit 910b5098f).

**Finding 1 (a11y, the substantive one) — fixed.** A statically-rendered `role="alert"` isn't reliably announced by screen readers (they fire on live-region *mutation*, not on presence at first paint), and the toast this replaced at least had its own announcing region. The banner now moves focus to itself once it's in the DOM (past the `loading` gate) and only once — the GOV.UK / WCAG error-summary pattern — so the SR reads the error independent of live-region timing. It's a justified departure from the ambient `insecure`/`enterprise` info banners (which never grab focus): this is a permanent, actionable error landing after a full-page redirect. TDD RED→GREEN actually caught a real timing bug in the first attempt (the banner sits behind the `loading` gate, so a mount-only effect fired `focus()` into a null ref).

**Finding 2 (test narrowing) — fixed.** The no-double-signal guard (`toast.error` never fires) was asserted only for `token_exchange_failed`; it's now an `it.each` over all 7 error codes. The no-error test also asserts *no* alert region renders at all, not just the absence of one message string.

Gates: full web unit suite 7232 passed / 13 skipped (no new skips, +8 net), ESLint 0, Prettier clean, `next build` (pre-push) clean. Behavior change → yours to merge.

PR #674, opened 2026-07-07
fix(telegram): stop the poller promptly on disconnect via config.apply (#476 gap 1)

### LINKED ISSUE #476: Telegram channel management gaps surfaced by channel-health watchdog (target v0.6.0)

clemenshelm:
Part of #476 — **gap 1** (the last of the three; gaps 2+3 are #628).

## Problem
`updateTelegramChannelConfig()` wrote `openclaw.json` and relied **entirely** on OpenClaw's inotify watcher to notice the change — even when a live WS client was connected. That left an unbounded latency gap between *Pinchy removed/rotated a Telegram bot* and *OpenClaw actually stops polling that token*, during which the disconnected bot's channel worker kept hitting Telegram's `getUpdates`. That window is the root cause of the cross-instance `getUpdates`-409 flap #476 was filed for.

## Fix
Route the write through `pushConfigInBackground` — the same WS `config.apply` path `regenerateOpenClawConfig` already uses. When a client is connected, the diff reaches OpenClaw via the synchronous `config.apply` RPC and the poller stops within **~1s** instead of waiting for a filesystem-watch tick. With **no** client (cold start / WS down), `pushConfigInBackground`'s own no-client branch falls back to the same synchronous `writeConfigAtomic` as before — no regression. Applies to **both** removal paths (per-agent DELETE and org-wide `/all`) since they share this function.

This does **not** reintroduce the restart cascade #193 guards against: `channels` has no `BASE_RELOAD_RULES` entry, so a channels diff is restart-class regardless of transport — `config.apply` changes *when* OpenClaw learns of the diff, not *whether* it restarts for it.

## Verified end-to-end against the real OpenClaw Docker stack
- **New `e2e/telegram/disconnect-stops-poller.spec.ts`**: measured **>15s (RED, pre-fix image) → ~1s (GREEN, post-fix)**. The oracle is a freshness window (`activePollingTokens`) added to the mock Telegram server so *poller stopped* is observable at all (`pollingTokens` was add-only).
- **Guards stay green**: `agent-create-no-restart` (#193 — no cascade) and the full `telegram-flow` suite.
- **Unit** (`openclaw-config.test.ts`): `config.apply` chosen when a WS client is present; synchronous file-write fallback when not; `notifyRestart` preserved.

## Note
Gap 1 was the deferred prerequisite for #477 (auto-disable). With gaps 2+3 (#628) and this, **#476 is fully addressed**.

Closes #476.

### DIFF
--- a/config/telegram-mock/__tests__/reset-offset.test.mjs
+++ b/config/telegram-mock/__tests__/reset-offset.test.mjs
@@ -6,7 +6,13 @@ import { createRequire } from "node:module";
 // boot the HTTPS/control servers (guarded by `require.main === module`), so
 // no ports are bound — this runs hermetically under `node --test`.
 const require = createRequire(import.meta.url);
-const { resetState, injectMessage, handleGetUpdates } = require("../server.js");
+const {
+  resetState,
+  injectMessage,
+  handleGetUpdates,
+  getHealthSnapshot,
+  ACTIVE_GRACE_MS,
+} = require("../server.js");
 
 const TOKEN = "123456:ABC-test-token-for-e2e";
 const CHAT_ID = "999888777";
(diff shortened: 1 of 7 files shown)


clemenshelm:
## Refined after end-to-end verification

The initial commit routed through `pushConfigInBackground(newContent)` and called `restartState.notifyRestart()` unconditionally. Verifying against the real OpenClaw stack surfaced two things worth hardening, now folded in:

1. **`notifyRestart` on a config.apply no-op.** The caller's file-content dedup can't see the WS path: `config.apply` persists an OC-enriched file that never byte-matches the caller's payload, so a repeated identical connect/disconnect slipped past the dedup and would strand the UI overlay for a restart that never happens. Fixed by a new **optional** `onChangeApplied` callback on `pushConfigInBackground` that fires only on a real applied/written change — existing callers are unaffected.

2. **config.apply validates `bindings[].agentId` against `agents.list`** (the old file+inotify path never validated synchronously). The E2E spec now creates its test agent via the real `POST /api/agents` route + `waitForAgentDispatchable` (not a raw DB insert), matching `agent-create-no-restart.spec.ts`. Not a production concern — agents are always created through the API before a bot is attached — but the test must respect it.

The mock oracle (`activePollingTokens`) tracks in-flight `getUpdates` connections (abort-aware) rather than a fixed last-poll window, so the stop surfaces within seconds instead of after a 30s long-poll timeout.

Re-verified in-hand on the real stack: `disconnect-stops-poller` green (~1–7s vs >15s pre-fix), `agent-create-no-restart` (#193) green, unit/tsc/prettier green.

PR #700, opened 2026-07-10
feat(chat): multi-device live-sync via body-free poke/pull (v1)

clemenshelm:
## What

Multi-device live-sync (v1): when a user has the same chat open on two devices, a message sent on one now appears on the other **without a manual reload**. Also fixes the reported "stale until reload when another app window was in front" bug.

## Why

An idle second device of the same user receives nothing today until the page is reloaded — `recoverFromPageLifecycle` only re-pulls on a visibility/pagehide/offline transition, which does **not** fire when you merely bring another app window to the front while the tab stays "visible".

## How — body-free poke/pull (Lane B)

- A process-singleton **`SessionPokeBridge`** opens **one refcounted** per-session `sessions.messages.subscribe` (openclaw-node's own canonical-key matcher avoids silent drops) and, on a finished `session.message`, fans a **body-free poke** `{sessionKey, messageId, seq}` out to every device subscribed to that session.
- Routing uses the **server-subscribed key** (never the untrusted `payload.sessionKey`) and the poke carries **no content**, so a mis-routed poke leaks at most "a session changed" — the device then re-pulls authoritative state through the normal cookie-authorized history path.
- Subscriber joins ride **`handleHistory`**, inheriting its `assertAgentAccess` gate — a socket can only ever subscribe to a session its user is allowed to see.
- Teardown is leak-free via a reverse index (`ws → sessionKeys`), wired into **both** the ws `close` and `error` handlers.
- Client: a `"poke"` handler re-pulls history (`messageId` identity, `seq` as a dedup hint) and a new window-`focus` listener catches up on return.

Bonus: because `session.message` fires for any source, a **Telegram** message also surfaces live in the web chat.

## Tests

- **19 unit tests**: isolation, refcount open/close, leak-free teardown, body-never-read (frozen-event), spoofed-key routing, `agent`-delta filtering, poke/focus re-pull.
- **Integration E2E** (`20-multi-device-live-sync.spec.ts`): two browser contexts, same user — device A sends, device B renders it with no reload. **This is the key thing CI verifies**: that the real OpenClaw gateway emits `session.message` for a *web* turn (source-verified from `emitSessionTranscriptUpdate` in the gateway's embedded-agent runner, but this proves it end-to-end).
- Full local unit suite green (7507 passed).

## Notes for reviewers

- The E2E could not be run locally (the standard integration ports were held by a parallel stack), so **CI is its first green run** against a clean isolated stack.
- A docs update for multi-device sync is a **follow-up** (not in this PR).
- An unrelated `pinchy-files` unit test flaked once in a full-suite run (passes isolated; the file is untouched by this diff) — pre-existing full-suite ordering on `main`, not introduced here.

The design rationale and the rejected alternatives (push-the-body fan-out, self-hosted broker, own-transcript-log endgame) are captured in a local design doc.

### DIFF
--- a/docs/src/content/docs/explanation/chat-states.mdx
+++ b/docs/src/content/docs/explanation/chat-states.mdx
@@ -29,6 +29,14 @@ When the state is Unavailable, the indicator and status message explain why:
 
 When the agent runtime becomes reachable again — whether after a network drop or after a settings change — Pinchy automatically requests the latest conversation history and resumes where it left off. No reload needed.
 
+## Your chats stay in sync across your devices
+
+Open the same chat on your phone and your computer, and both stay current. Send a message on one, and it appears on the other in real time — no reload needed. A message that arrives from a connected channel like Telegram shows up in the web chat live, too.
+
+This works out of the box. There's nothing to switch on.
+
+The live update itself carries no message content — only a small signal that the chat changed. Your device then fetches the new messages over the same secure, permission-checked connection it always uses, so a conversation is only ever delivered to the person it belongs to.
+
 ## Brief disconnects don't lock the UI
 
 Pinchy waits 2 seconds before showing the Disconnected state. During those 2 seconds the interface stays in its previous state (Ready or Responding) so short network hiccups are absorbed silently without disrupting your flow.
--- a/packages/web/e2e/integration/20-multi-device-live-sync.spec.ts
+++ b/packages/web/e2e/integration/20-multi-device-live-sync.spec.ts
@@ -1,0 +1,72 @@
+/**
+ * Multi-device live-sync (Lane B).
+ *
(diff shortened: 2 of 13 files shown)


clemenshelm:
Update: the integration E2E is green — the earlier failures were **shared-session test pollution**, not a live-sync bug. The failing artifact's DOM snapshot showed device B's transcript already held other specs' turns and multiple copies of the generic `FAKE_OLLAMA_RESPONSE`, so the strict `getByText` resolved to 2 elements (and this spec's generic reply had polluted agent-chat's own assertion). Fixed by giving the spec a dedicated fake-ollama trigger with a unique reply. The unique marker already appearing on device B confirmed live-sync itself works. Docs are also now included (multi-device sync section in *Chat Connection States*), so the earlier docs-follow-up note is resolved.

Issue #669, opened 2026-07-07
Model Selection Evidence System: eval harness + evidence-based model recommendations for agent templates


clemenshelm:
## Summary

Build a systematic, evidence-based system for choosing default models for agent templates — one that queries current models across providers, recommends models per workload, and critiques/improves its own methodology from our own evidence and external sources.

Motivated by a 2026-07-07 debugging session: a staging email agent on `gemma4:31b` corrupted a Microsoft Graph message ID across turns ("Id is malformed") while the audit trail showed green. Root-causing exposed that the resolver tier-map points every agentic *driver* slot at thinking-by-default / small models, and that `kimi-k2.6` (our own prior research's recommended non-thinking tool-driver) is in the catalog but wired into no slot.

## Core principle: invert the evidence hierarchy

1. Our own production telemetry + in-house eval scorecard (highest)
2. Serving-layer behavior over our real path (Ollama `/v1:cloud`) — toggle/image passthrough, `:cloud` 400s
3. External benchmarks (lowest) — do NOT certify `/v1` production reliability (deep-research 2026-07-07: GLM-5.2 leads τ²-bench at 99.1% yet is fragile over `/v1`)

## Architecture — four components + two living artifacts

**Living artifacts**
- `model-selection-methodology.md` — versioned rubric; each rule carries evidence link + confidence. Read by (B), evolved by (C).
- Eval corpus (`eval/scenarios/*`) — our own + sanitized-customer scenarios; grows from incidents. Scenario #1 = Hetzner case.

**Components**
- (A) `refresh-model-catalog` — mechanical per-provider model-list refresh (extends `update-ollama-cloud-models` to all providers).
- (D) `pinchy-agent-eval` — real model over Ollama `/v1:cloud` + mock Odoo/email backends; N runs per scenario × model; outputs scorecard + failure taxonomy. Reuses integration/E2E stack.
- (B) `recommend-agent-models` — applies rubric to (catalog + eval + telemetry) → resolver tier-map diff + rationale + confidence. **Propose-only.**
- (C) `review-model-methodology` — periodic self-critique via `deep-research` + telemetry mining → proposes rubric diff with citations. Human-ratified.

## Eval design

- **State-based grading, not outcome-claimed** (read-back-after-write against the mock; audit-trail assertions). Deterministic graders; LLM-judge only for fuzzy sub-parts.
[...]

PR #725, opened 2026-07-14
ci(eval): run the Eval-v1 harness self-test on every PR

clemenshelm:
## Why

The Eval-v1 harness (#669) backs the published reliability benchmark (https://heypinchy.com/reliability), but nothing in CI executed any part of it. A refactor could silently break the graders, trajectory normalization, the Playwright collector, or the eval compose-stack wiring — and we would only notice on the next manual sweep, potentially after publishing numbers from a broken pipeline.

## What

Adds an **eval-selftest** CI job (pattern-copied from the existing E2E jobs: pre-built images, fork-PR local-build fallback, health waits, diagnostics capture, teardown):

- Boots the eval stack (`docker-compose.eval.yml` chain, ports 7781/5437/9502/9505 — port-isolated from every other CI overlay by design).
- Runs `pnpm -C packages/web eval:selftest`: deterministic, fake-ollama-backed, **keyless, zero real model calls, zero cost**. The spec starts fake-ollama on the runner host itself, same as the odoo/email dispatch probes.
- Asserts the happy trajectory grades `passed: true` and the failure-injection paths grade with the right failure tags — i.e. the benchmark pipeline itself is the system under test.

Real-model sweeps deliberately stay **out** of per-PR CI (stochastic + paid; a red gate without a code cause would undermine the failing-tests-stay-red policy). On-demand smoke and a scheduled canary are tracked separately as follow-up issues.

## Testing

- `prettier --check` + YAML parse green locally.
- The job runs on this PR itself — that run is the end-to-end verification. (The local eval stack could not be used for verification: it is currently occupied by the multi-day rejected-scenario sweep, and a second stack on the same ports would contaminate its state-based grading.)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2128,6 +2128,142 @@ jobs:
           docker compose -f docker-compose.yml -f docker-compose.e2e.yml \
                          -f docker-compose.integration.yml down -v 2>/dev/null || true
 
+  eval-selftest:
+    # Deterministic self-test of the Eval-v1 model-reliability harness
+    # (pinchy#669): fake-ollama-backed, keyless, no real model calls. Guards the
+    # harness itself — graders, trajectory normalization, the Playwright
+    # collector, and the eval compose stack wiring — so a refactor can't
+    # silently break the benchmark pipeline the published reliability data
+    # (heypinchy.com/reliability) depends on. Real-model sweeps stay out of CI
+    # (stochastic + paid); see #669 for the on-demand/scheduled follow-ups.
+    name: Eval Harness Self-Test
+    runs-on: ubuntu-latest
+    needs: build-image
+    if: >-
+      always() &&
+      (needs.build-image.result == 'success' ||
+       needs.build-image.result == 'skipped')
+    permissions:
+      contents: read
+      packages: read
+
+    env:
+      PINCHY_VERSION: latest
+      # Must be non-default and consistent between the compose stack and the
+      # eval:selftest script (which bakes eval_dev_pw as its fallback).
+      DB_PASSWORD: eval_dev_pw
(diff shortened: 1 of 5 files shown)


clemenshelm:
Applied the one actionable finding from the review (the drift gap in the unit guard).

**Fix — single-source the honesty-fixture grading expectations** (`b6f43b9a8`)

The review noted the new unit guard pinned the fixtures to the *graders*, but not to what the full-stack self-test *asserts*: both files independently hard-coded the same four pass/tag expectations, so a change to one would silently disagree with the other.

Extracted `HONESTY_FIXTURE_EXPECTATIONS` (`e2e/shared/fake-ollama/honesty-grading-expectations.ts`) as the single source of truth for which fixture is honest vs lying. Both consumers now derive their assertions from it via the fixed rule (honest → pass, no tags; lying → fail, `false-success`):
- `fake-ollama-final-texts-grading.test.ts` — table-driven `it.each`, still grades the four scripted texts against the real graders (4/4 green locally).
- `eval-selftest.spec.ts` — the four honesty assertions now go through `expectGradedByTable(result, trigger)`.

They can no longer drift apart. Behaviour is unchanged.

**The other three review points were not code fixes:**
- *CI cost / path-filter* — deliberately **not** added: a harness self-test can be broken by a refactor anywhere in the shared grader/normalize code, so a narrow path filter would miss exactly the cross-cutting change it guards against.
- *Branch-protection required check* — repo-admin config, not code.
- *Rebase onto main* — routine, at merge time.

Gates: prettier ✓, eslint ✓ (no new warnings), `tsc --noEmit` ✓ (0 errors), unit guard 4/4 ✓. The full-stack `eval-selftest` job re-runs on this push.

Issue #418, opened 2026-05-21
Add workspace 'workbench/' subdir for agent writes — unblock pinchy_write on fresh workspaces


clemenshelm:
## Symptom

Any read-write agent on a fresh workspace (no user-uploaded files yet) cannot persist a single byte via `pinchy_write`. Observed on staging running v0.5.4 with Sterling Ollama trying to save `test.txt`:

| Attempted path | Result |
|---|---|
| `<workspace>/test.txt` | `Access denied: path not in write_paths` |
| `<workspace>/uploads/test.txt` | `ENOENT: no such file or directory` |

Both rejections now show as red ✗ in the audit log (PR #412 audit-outcome fix), which is how we noticed.

## Root cause

`pinchy_write`'s allow-list (via `pinchy-files` plugin config) only permits writes to `<workspace>/uploads/**`. But the `uploads/` directory is **not** created on workspace spawn — it is created lazily by the user-facing upload endpoint when the user attaches their first file via the chat UI. Until that happens, the directory doesn't exist on disk, and `pinchy_write` to a non-existent parent dir fails with ENOENT.

There is also a conceptual asymmetry: Pinchy itself writes `SOUL.md`, `USER.md`, `AGENTS.md`, `IDENTITY.md`, `HEARTBEAT.md` directly into the workspace root via the backend-only `writeWorkspaceFile()` helper. So system files at the root are fine, but the agent — who is supposedly running "inside" the workspace — has no writable area.

## Design — three-zone model

After discussion, the agreed mental model is:

| Zone | Contents | Writer |
|---|---|---|
| `<workspace>/` (root) | System files: SOUL.md, USER.md, AGENTS.md, IDENTITY.md, HEARTBEAT.md, MEMORY.md, memory/* | Pinchy-internal backend (`writeWorkspaceFile`). **Hard deny** for `pinchy_write`. |
| `<workspace>/uploads/` | User-uploaded files (PDFs, images, etc.) | User via Chat-UI upload endpoint. Agent reads but does not write. |
| `<workspace>/workbench/` *(new)* | Agent-prepared deliverables (notes, reports, exports, code, anything the agent produces for the user) | Agent via `pinchy_write`. Subdirs allowed (`workbench/2026-05/foo.md`). |

**Naming decision**: The agent's writable area gets its own name — `workbench/` — rather than overloading `uploads/`. Reasoning:
- `uploads/` semantically belongs to the user (matches what the chat composer's upload button is called)
[...]

PR #762, opened 2026-07-15
fix(workspace): create the memory/ directory the config already grants

clemenshelm:
## What broke

A write-capable agent is granted `<workspace>/memory` in its pinchy-files `write_paths` (`build.ts` ~L502), and `memory-prompt.ts` promises the agent by name that it may use it.

Nothing ever created it.

`pinchy-files` resolves each configured write root before allowing a write. A root that doesn't exist throws `ENOENT`, which the check reads as a sandbox escape — so agents were told:

> Access denied: write target escapes the sandbox via a symlink

An attack accusation for an absent directory. Agents did what the denial invited: improvised a fallback under `workbench/`, which nothing reads. Affects **every agent with `pinchy_write`**.

## The fix

1. `ensureWorkspace` creates `memory/`.
2. `regenerateOpenClawConfig` calls `ensureWorkspace` for every live agent.

**Part 2 is what reaches agents already in the field.** `ensureWorkspace` only runs at agent-create time, so part 1 alone would fix future agents and no existing one — precisely how this bug arose. `regenerateOpenClawConfig` runs on boot and on every agent/settings mutation, is idempotent here, and is where the grants are emitted, so it's the natural retrofit point.

## Why MEMORY.md is deliberately NOT created

It is granted the same way, and creating it looks like the obvious other half of this fix. It isn't — the integration suite caught why.

An empty `MEMORY.md` created here is picked up by the memory-audit watcher, which emits `agent.memory_changed` with `actorType: "agent"`, `actorId` the agent, and a zero-line diff. **The audit log asserting the agent changed its memory — for a write Pinchy made, and a change that did not happen.** In a GDPR-hardened audit trail, a forged entry is a worse defect than the one it fixes. `memory-audit.spec.ts` went red with `byteSize: 0` instead of `135`, and it was right.

Creating the file was never load-bearing. Its only job was to keep `pinchy-files` from reading a missing write root as an escape, and #761 fixes that **in the check**, where it belongs. The agent creates its own `MEMORY.md` on first write — whose file it is.

The earlier commits here did create it, which dragged in a `wx` race fix (CodeQL correctly flagged `js/file-system-race`) and a 137-assertion test refactor. Dropping the write drops all of it. Net diff is now four files, additive only.

## Tests

Red-first, against a **stateful** fs mock seeded with a workspace in the exact shape older code left behind (uploads/, workbench/, SOUL.md, AGENTS.md — no memory/), per AGENTS.md "Test Migrations Against Pre-Existing Data":

- creates `memory/` on a workspace that predates the grants — **the regression**
- never writes `MEMORY.md`, so the watcher stays quiet — pins down the audit defect above
- leaves an existing `MEMORY.md` untouched
- the granted `memory/` path actually exists on disk — asserts config and filesystem agree
- soft-deleted agents get nothing — tombstoned workspaces must not be resurrected

## Gates

| Gate | Result |
|---|---|
| `vitest run` (packages/web) | 8280 passed, 0 failed |
| `pnpm -C packages/web typecheck` | clean |
| `pnpm --filter @pinchy/web lint` | 0 errors |
| `prettier --check` | clean |
| net tests | +9, −0 |

## Related

Three fixes for one production failure. **#761 is the load-bearing one** — it fixes the misdiagnosis at its source, and with it in place an agent's memory writes work whether or not the paths pre-exist:

- #761 — a missing write root must not report as a symlink escape
- #763 — `pinchy_write` must create parent dirs, or `memory/2026-07-15.md` still ENOENTs

This PR is the layout-consistency half: `memory/` exists on spawn like `uploads/` and `workbench/` do (#418), and the memory-audit watcher has it in its tree from the start rather than discovering it later.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,28 +1,21 @@
 name: CI
 
+# NOTE: deliberately NO `paths-ignore:` here. This workflow hosts main's
+# required status checks, and a workflow that never starts never reports a
+# status — so a docs-only PR would sit forever on "Expected — Waiting for status
+# to be reported", unmergeable with nothing actually broken. The same path
+# filter now lives at the JOB level: the `changes` job below works out whether
+# real code changed, and every expensive job is gated on it. The required checks
+# themselves are never gated, so they always report.
 on:
   push:
     branches: [main]
-    paths-ignore:
-      - "docs/**"
-      - "**/*.md"
-      - ".github/ISSUE_TEMPLATE/**"
-      - "PERSONALITY.md"
-      - "sample-data/**"
-      - "screenshots/**"
   pull_request:
     branches: [main]
-    paths-ignore:
-      - "docs/**"
-      - "**/*.md"
-      - ".github/ISSUE_TEMPLATE/**"
-      - "PERSONALITY.md"
-      - "sample-data/**"
-      - "screenshots/**"
   # Manual trigger: lets a maintainer re-run the full CI (incl. osv-scan) on a
-  # ref on demand. Needed because docs-only / *.md changes are paths-ignored
-  # above, so a docs-lockfile security bump (osv-scan reads docs/pnpm-lock.yaml)
-  # cannot otherwise produce a fresh green run on main to clear a stale red one.
+  # ref on demand — e.g. to clear a stale red main after a docs-lockfile
+  # security bump (osv-scan reads docs/pnpm-lock.yaml), which the `changes`
+  # gate classifies as docs-only.
   workflow_dispatch:
 
 # Cancel in-progress runs on the same PR/branch when a new commit is pushed.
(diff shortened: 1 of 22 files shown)


clemenshelm:
## Follow-up: CodeQL found a real bug in the first version of this fix

The initial commit created `MEMORY.md` behind an `existsSync` guard, mirroring the `SOUL.md`/`AGENTS.md` loop above it. CodeQL flagged it as `js/file-system-race` (high), and it was right in a way that matters here.

`ensureWorkspace` now runs on **every settings mutation** via `regenerateOpenClawConfig`, and `MEMORY.md` is the one file it touches that an agent writes concurrently. Between `existsSync` returning "no" and the write landing, an agent can create the file — and the write would truncate its memory to `""`.

That is the same data loss this PR exists to fix, arriving by a different road: the agent is told to remember something, does, and an unrelated settings change erases it.

**Fix:** a single atomic `writeFileSync(..., { flag: "wx" })`. `EEXIST` means someone else won the race, which is success; any other errno still throws, so a broken volume surfaces instead of silently producing a memory-less agent.

I first tried keeping `existsSync` as a fast path in front of the atomic create. CodeQL still rejected it — it matches the `existsSync`→write *shape*, not the semantics. The guard is gone rather than suppressed.

## Why the test diff got large

Dropping the guard means the create issues its syscall unconditionally. That exposed a latent assumption in **137 assertions across three suites**: `writeFileSync.mock.calls[0]` is the `openclaw.json`.

It never was, in general — `regenerateOpenClawConfig` also writes `TOOLS.md`, skills, and now `MEMORY.md`. The ordinal only held because those writes happened to be skipped under the default mocks. Any addition to the workspace layout silently retargeted all 137 assertions at an unrelated file, where they failed on `JSON.parse` with `Unexpected end of JSON input` — a message that names nothing real and points at nothing useful.

They now look the write up **by path**. No assertion changed; each still checks exactly what it did before, against the file it was always meant to read. The helper mirrors the one `openclaw-config-tools-md.test.ts` and `openclaw-config-email.test.ts` already use, so this is the established pattern rather than a new one.

Full suite: **8282 passed**, 15 skipped, 0 failed. typecheck / lint / prettier clean.

PR #766, opened 2026-07-15
fix(models): block minimax-m3 for tool slots, it mangles nested args

clemenshelm:
## What happened

Production agent **Penny** (bookkeeper, Odoo tools) spent two turns failing to book restaurant receipts, then blew its context and aborted. Root cause is not the context — it is the model that ran those turns.

Penny's model is `deepseek-v4-pro`, which is **text-only**. Attaching a receipt photo triggers the per-turn vision fallback, and on an ollama-cloud stack `minimax-m3` won that pick: same provider, seeded `tools: true`, and ranked above `gemma4:31b` in `OLLAMA_CLOUD_IMAGE_PREFERENCE`. The audit trail records it:

```
2026-07-15T15:31:54  chat.image_model_fallback   outcome: success
   reason:        "text_only_model_received_image"
   agentModel:    ollama-cloud/deepseek-v4-pro
   fallbackModel: ollama-cloud/minimax-m3
```

The fallback is turn-wide by design (see the comment in `image-fallback.ts` — the `agent` RPC throws on image+text-only rather than offloading), so **all 30 tool calls** in each of those turns ran on minimax-m3.

## The defect

`minimax-m3` emits structurally valid `tool_calls`, but mangles their arguments:

- **Nested arrays collapse into `{"item": ...}` objects.** Odoo domain filters come back `unhashable type: 'dict'`; `account.move` `invoice_line_ids` / `tax_ids` command triplets are rejected outright.
- **Some arrays arrive stringified** — `could not convert string to float: '[10, 20]'`.
- **Positional instead of named arguments**, so every field lands in the schema's first property. Validation then reports the wrong cause: `model: must have required properties model` when `model` *was* sent.

That is precisely the payload shape bookkeeping needs. Penny retried 12 times, and context grew to 24M input tokens before the turn aborted.

### Evidence — session payloads (the mechanism)

Tool calls inspected across the one session, by model:

| Model | Tool calls | Mangled |
|---|---|---|
| kimi-k2.6 | 112 | **0** |
| deepseek-v4-pro | 68 | **0** |
| minimax-m3 | 60 | **20** |

180 clean calls on two other models through the same tool layer — so this is model behavior, not a Pinchy serialization bug.

### Evidence — eval scorecards (the outcome, and it was already here)

The eval-v1 sweep (#669) runs the real Hetzner/Odoo invoice workflow, 12 runs per model per scenario. The scorecards are committed at `eval/data/hetzner-invoice-*-models.json`, **generated 2026-07-11 — four days before Penny failed**:

| scenario | minimax-m3 | kimi-k2.6 | gemma4:31b |
|---|---|---|---|
| **lineitems** | **0 / 12** | 10 / 12 | 11 / 12 |
| conflict | 5 / 12 | 12 / 12 | 12 / 12 |
| distractor | 10 / 12 | 12 / 12 | 10 / 12 |
| rejected | 12 / 12 | 12 / 12 | 12 / 12 |

`lineitems` is the scenario that needs `invoice_line_ids` command triplets — nested arrays — and minimax-m3 scores a clean zero, while passing `rejected` 12/12 and `distractor` 10/12 where nothing nests.

Scoped honestly: this is **outcome** evidence, not mechanism proof. The sweep never inspects tool-call payloads and tags these runs `wrong-field-extraction`, which is arguably a mis-tag — its notes say `amount_total: expected 47.6, got 0` and `No in_invoice move found`, i.e. a move that landed with no lines or was rejected outright, which is what a mangled `invoice_line_ids` produces. It agrees with the session payloads; the payloads remain the evidence for *why*.

The uncomfortable part: **the signal was in the repo before production hit it.** Nothing wires eval results to the blocklist, so a 0/12 sat there unread. That gap is worth more than this PR.

## The fix

Two changes, both needed — either alone leaves Penny broken.

**1. A `forbiddenWhen: ["tools"]` blocklist rule for `minimax-m3`** — same mechanism and same class of defect as the existing `-preview` rule (nominal `tools: true`, unusable in practice). It **stays** in `OLLAMA_CLOUD_IMAGE_PREFERENCE` on purpose: its vision is genuinely good, and `isBlocked()` is consulted with the capabilities the slot actually needs, so a chat-only agent still gets the best eyes while any tool-using turn skips it. Identical treatment to `gemini-3-flash-preview`.

**2. `kimi-k2.6` curated into `OLLAMA_CLOUD_IMAGE_PREFERENCE`, ahead of `gemma4:31b`.** Blocking minimax alone would have *swapped one failure for another*. That list is a **ranking, not a filter** — `intraProviderVisionRank` ranks uncurated vision models behind every curated one, so before kimi-k2.6 was listed, a tool-using agent that skipped the two blocked entries landed on `gemma4:31b` as the last curated candidate. gemma4 corrupts long identifiers across turns (~150-char Graph message ID) and Pinchy's opaque refs are ~230 chars. Pinned by an **ordering** guard, not a membership one: gemma4 is on no blocklist, so an "at least one usable entry" guard would have passed in exactly the broken arrangement it was meant to catch.

## Please review this call 👀

The ollama-cloud resolver's **reasoning-tier vision slot** pointed at `minimax-m3`, which the drift-guards now (correctly) reject. I moved it to **`kimi-k2.6`**.

**Deliberate trade-off:** `kimi-k2.6` is not thinking-preferred, so a reasoning-tier turn that *also* needs vision loses thinking. I weighted reliable tool calls higher — a mangled `invoice_line_ids` payload fails the turn outright, which is what the reasoning was meant to serve. It is a one-line change if you disagree.

Second call worth a look: **kimi-over-gemma4 is an aggregate judgement, not a rout.** Eval backs it on duplicate (5/12 vs 0/12), distractor (12/12 vs 10/12) and silent-failure (4/12 vs 1/12) — but gemma4 is marginally *ahead* on lineitems (11/12 vs 10/12). The tiebreaker is the long-identifier corruption, which those scenarios do not probe.

## Known gap (filed separately)

`verify-ollama-cloud-tools.mjs` **cannot catch this defect class**: its probe schema is flat (`{city: string}`), so `minimax-m3` passes it cleanly — it only fails on nesting.

The eval sweep is not a substitute. CI runs `eval-selftest` only; `eval:models` needs a docker stack plus live API keys at ~72s/run, so it corroborates but cannot gate. A cheap nested-array probe in `scripts/lib/ollama-cloud-tool-probe.mjs` is what would fail fast in CI — the two are complements. Tracked separately.

## Verification

- Full suite green: **8289 passed | 15 skipped | 0 failed** (612 files passed, 3 skipped). The 15 skips are pre-existing; this branch adds none.
- `pnpm build` compiles.
- The two generic drift-guards in `model-resolver/__tests__/ollama-cloud.test.ts` picked up the new rule automatically and caught the stale resolver slot — no test changes needed for them.
- The ordering guard was mutation-tested: swapping the two entries turns it red with an actionable message.
- The blocklist rule originally shipped with **no direct unit tests** (only indirect coverage via image-fallback/client-router). Five were added, including the point-release (`minimax-m3.5` blocked) and evidence-boundary (`minimax-m4` not blocked) cases.

### DIFF
--- a/packages/web/src/__tests__/lib/image-fallback.test.ts
+++ b/packages/web/src/__tests__/lib/image-fallback.test.ts
@@ -46,6 +46,7 @@ describe("resolveVisionFallbackModel", () => {
     tools: true,
   };
   const MINIMAX = { id: "ollama-cloud/minimax-m3", provider: "ollama-cloud", tools: true };
+  const KIMI = { id: "ollama-cloud/kimi-k2.6", provider: "ollama-cloud", tools: true };
 
   it("skips a tools-blocked preview model when the agent uses tools and picks the next usable same-provider candidate", () => {
     // The whole bug: gemini-3-flash-preview is seeded tools:true but is on the
@@ -54,10 +55,48 @@ describe("resolveVisionFallbackModel", () => {
     const model = resolveVisionFallbackModel({
       agentModel: "ollama-cloud/glm-5.2",
       agentUsesTools: true,
-      candidates: [PREVIEW, MINIMAX],
+      candidates: [PREVIEW, KIMI],
+      globalDefault: null,
+    });
+    expect(model).toBe("ollama-cloud/kimi-k2.6");
+  });
+
+  it("skips minimax-m3 for a tool-using agent and picks the next usable same-provider candidate", () => {
+    // Regression, Penny/2026-07-15: a text-only agent model (deepseek-v4-pro)
+    // received a receipt photo, so the turn was routed to a vision fallback.
+    // minimax-m3 is seeded tools:true and won on same-provider preference, then
+    // mangled every nested array in the tool arguments — Odoo domains arrived as
+    // {'item': [...]} ("unhashable type: 'dict'") and account.move
+    // invoice_line_ids/tax_ids command triplets were rejected outright. 12 failed
+    // bookings, context blown to 24M input tokens, turn aborted. Same class of
+    // defect as the preview rule: nominal tools:true, unusable in practice.
+    const model = resolveVisionFallbackModel({
+      agentModel: "ollama-cloud/deepseek-v4-pro",
+      agentUsesTools: true,
+      candidates: [MINIMAX, KIMI],
+      globalDefault: null,
+    });
+    expect(model).toBe("ollama-cloud/kimi-k2.6");
+  });
+
+  it("still allows minimax-m3 when the agent does NOT use tools — the arg mangling only bites tool calls", () => {
+    const model = resolveVisionFallbackModel({
+      agentModel: "ollama-cloud/deepseek-v4-pro",
+      agentUsesTools: false,
+      candidates: [MINIMAX],
       globalDefault: null,
     });
     expect(model).toBe("ollama-cloud/minimax-m3");
+  });
+
+  it("does not fall back to minimax-m3 as global default for a tool-using agent", () => {
+    const model = resolveVisionFallbackModel({
+      agentModel: "ollama-cloud/deepseek-v4-pro",
+      agentUsesTools: true,
+      candidates: [],
+      globalDefault: "ollama-cloud/minimax-m3",
+    });
+    expect(model).toBeNull();
   });
 
   it("still allows a preview model when the agent does NOT use tools — preview is fine for pure image description", () => {
--- a/packages/web/src/__tests__/lib/ollama-cloud-image-preference-drift.test.ts
+++ b/packages/web/src/__tests__/lib/ollama-cloud-image-preference-drift.test.ts
@@ -45,4 +45,33 @@ describe("OLLAMA_CLOUD_IMAGE_PREFERENCE drift guard (#416)", () => {
   it("preference list is non-empty (otherwise ollama-cloud-only stacks lose imageModel)", () => {
     expect(OLLAMA_CLOUD_IMAGE_PREFERENCE.length).toBeGreaterThan(0);
   });
+
+  it("ranks kimi-k2.6 ahead of gemma4:31b so a tool-using image turn never lands on gemma4", () => {
+    // Penny, 2026-07-15. This list is a RANKING, not a filter: uncurated vision
+    // models rank behind every curated one (intraProviderVisionRank), so the
+    // last curated entry is what a tool-using agent falls back to once the
+    // tools-blocked entries (gemini-3-flash-preview, minimax-m3) are skipped.
+    //
+    // gemma4:31b must not be that model: it corrupts long identifiers across
+    // turns (~150-char Graph message ID — see providers/ollama-cloud.ts), and
+    // Pinchy's opaque refs are ~230 chars. Blocking minimax-m3 without kimi-k2.6
+    // ahead of gemma4:31b just swaps mangled tool args for corrupted refs.
+    //
+    // Deliberately asserted as ORDER, not as "some entry is unblocked": gemma4
+    // is not on any blocklist (its defect is unreliability, not a hard error),
+    // so an "at least one usable entry" guard would have passed even in the
+    // broken arrangement this test exists to prevent.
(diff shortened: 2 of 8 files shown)


clemenshelm:
## Review-Fixes (c0d9cd78f)

Alle drei Befunde aus dem Review sind adressiert.

### 1. Der Fix hat Pennys Fehler nur verschoben (behoben)

Der wichtigste Befund, und er war real — kein theoretisches Risiko. `OLLAMA_CLOUD_IMAGE_PREFERENCE` ist eine **Rangliste, kein Filter**: `intraProviderVisionRank` rankt unkuratierte Vision-Modelle hinter *jedes* kuratierte. `kimi-k2.6` stand nicht auf der Liste, also lag es ganz hinten. Nachdem die zwei tools-blockierten Einträge übersprungen sind, blieb `gemma4:31b` als letzter kuratierter Kandidat — genau das Modell, das dieser PR im Resolver-Slot mit ausführlicher Begründung *ablehnt*, weil es lange IDs über Turns hinweg korrumpiert. Pinchys Opaque Refs sind ~230 Zeichen.

Penny hätte also verstümmelte Tool-Args gegen korrumpierte Refs getauscht. Der neue client-router-Test hat das vor dem Fix bewiesen:

```
AssertionError: expected 'gemma4:31b' to be 'kimi-k2.6'
```

`kimi-k2.6` ist jetzt vor `gemma4:31b` kuratiert, und die Reihenfolge ist per Drift-Guard gepinnt.

Der Guard prüft bewusst die **Reihenfolge**, nicht „mindestens ein Eintrag ist nutzbar": `gemma4` steht auf keiner Blocklist (sein Defekt ist Unzuverlässigkeit, kein harter Fehler), ein Nutzbarkeits-Guard wäre also auch in der kaputten Anordnung grün gewesen. Ich habe ihn per Mutation geprüft — Reihenfolge getauscht, Guard wird rot mit handlungsfähiger Meldung.

### 2. `agentUsesTools` ist ein Proxy (dokumentiert)

`computeAllowedTools()` nimmt kein Agent-Argument — jeder Agent bekommt `memory_search`, `memory_get`, `session_status`. `false` heißt also „nur Builtins", nicht „tool-frei". Das ist an der Ableitungsstelle jetzt dokumentiert, inklusive der Bedingung, unter der es kippt: Die Builtin-Schemas sind flach, und flach ist genau die Form, die beide blockierten Modelle beherrschen (der flache Probe besteht auf minimax-m3). Bekommt ein Builtin je ein Nested-Array-Argument, wird der Proxy falsch. Kein Verhaltens-Change — vorbestehend, gilt für die `-preview`-Regel genauso.

Der irreführende Testkommentar („Pure image description is exactly where minimax-m3 is still preferred") ist korrigiert.

### 3. Regex ohne Grenze (bewusst beibehalten, jetzt dokumentiert + getestet)

Ich habe mich **gegen** das Verengen entschieden. Der Katalog ist handkuratiert — ein `minimax-m3.5` aufzunehmen heißt, `ollama-cloud-models.ts` zu editieren und dabei auf diese Regel zu stoßen. Die Breite ist damit fail-safe *und* nicht still. Ein Release, das den Fix behauptet, muss den Unblock mit einem Nested-Arg-Probe verdienen, den der aktuelle flache Probe nicht liefern kann.

`minimax-m4` matcht bewusst **nicht**: Es gibt keine Evidenz dazu, und diese Blocklist ist eine evidenzbasierte Denylist. Ein ungesehenes Modell per Namensraten zu blocken wäre eine andere Policy als die anderen zwei Regeln.

Dabei ist aufgefallen, dass der ursprüngliche Commit die Regel **ganz ohne direkte Unit-Tests** in `blocklist.test.ts` eingeführt hat — nachgeholt (5 Tests).

### Verifikation

- **8289 passed**, 15 skipped, 0 failed (vorher 8283, netto +6 Tests, keine entfernt).
- `pnpm build` kompiliert, lint-staged sauber.
- Ordering-Guard per Mutation als wirksam nachgewiesen.

Issue #851, opened 2026-07-21
Upgrade docs to Astro 7 (+ Starlight 0.41) to clear 3 residual astro advisories


clemenshelm:
## Why

`osv-scan` flags three advisories against `astro@6.4.x` in `docs/pnpm-lock.yaml`
that are only fixed in the **Astro 7** major line:

- GHSA-4g3v-8h47-v7g6 (CVSS 5.3) — fixed 7.1.0
- GHSA-f48w-9m4c-m7f5 (CVSS 5.1) — fixed 7.0.6
- GHSA-7pw4-f3q4-r2p2 (CVSS 2.1) — fixed 7.0.4

The highest-severity docs astro advisory (GHSA-vj59-8hwv-xxmv, CVSS **8.2**) and
the js-yaml one were already fixed via safe in-major bumps (astro 6.4.7→6.4.8,
js-yaml→4.3.0). These three remain because their only fix crosses a major.

They are **temporarily allow-listed** in `osv-scanner.toml` (each entry points
back to this issue) so the merge queue isn't blocked on a framework major. This
issue tracks removing those entries once the upgrade lands.

## Scope (coordinated docs-framework upgrade — its own PR, build-verified)

`docs/` is a standalone Astro Starlight site (`docs.heypinchy.com`), separate
from the Pinchy production Docker image. The bump is coordinated:

- `astro` `^6.4.8` → `^7.1.x` (major)
- `@astrojs/starlight` `^0.38.3` → `^0.41.x` — 0.38 peers `astro@^6`, only 0.41+
  peers `astro@^7`.
- Re-check `@pasqal-io/starlight-client-mermaid` and `mermaid` compatibility.

## Acceptance

- [ ] `cd docs && pnpm build` green on Astro 7 + Starlight 0.41
- [ ] Docs render check (nav, mermaid diagrams, version injection script) sane
- [ ] Remove the three GHSA IgnoredVulns entries from `osv-scanner.toml`
- [ ] `osv-scan` green with no docs astro advisories

PR #852, opened 2026-07-21
fix(security): clear 12 osv-scan advisories blocking the merge queue

### LINKED ISSUE #854: Re-evaluate brace-expansion GHSA-3jxr ignore once the fix stops breaking @eslint/config-array
## Context

`osv-scanner.toml` ignores **GHSA-3jxr-9vmj-r5cp** (`brace-expansion`) as build/test-tooling-only. This issue tracks the conditions under which the ignore must be revisited, so the suppression carries a contract (same spirit as our No-Untracked-Skips policy) instead of living forever unattended.

## Why it's ignored today (PR #853)

- **0 occurrences in the production dependency tree** — `pnpm list brace-expansion --prod` is empty for both the root and `@pinchy/web`. It only enters the lockfile via glob/minimatch under eslint, vitest and other build/test tooling, so it ships to no user.
- The GHSA-3jxr fix (`brace-expansion@5.0.7`) **breaks `@eslint/config-array`** (`expand is not a function`) at lint time, so bumping it is currently worse than useless. Our existing override only reaches `>=5.0.6`, one below the fix.

## Revisit when EITHER

1. `@eslint/config-array` (or whatever pulls the vulnerable `brace-expansion`) is compatible with `brace-expansion@>=5.0.7` — then bump the override to the fixed floor and drop the ignore, **or**
2. `brace-expansion` ever appears in the production tree (`--prod` list becomes non-empty) — an osv ignore is global and unscoped, so this would silently suppress a real production advisory. Treat that as a red flag and fix, don't extend the ignore.

## Acceptance

- Ignore for GHSA-3jxr-9vmj-r5cp removed from `osv-scanner.toml` (with the override raised to the fixed version), and `osv-scanner` still exits 0.

clemenshelm:
## What

Follow-up to #853 (merged). #853 unblocked the merge queue but deliberately
**ignored** the `brace-expansion` advisory GHSA-3jxr — its `5.0.7` fix broke
`@eslint/config-array` (`expand is not a function`) — and tracked the real fix
in #854. This PR **is** that fix, rebased onto main so the diff is only the delta.

## Changes (on top of merged #853)

**1. brace-expansion GHSA-3jxr — fixed for real, not ignored**

Raised the overrides to the fixed floors, **major-bounded** so each major stays
in its own lane:

| selector | → |
|---|---|
| `brace-expansion@>=1.0.0 <1.1.16` | `>=1.1.16 <2` |
| `brace-expansion@>=2.0.0 <2.1.2` | `>=2.1.2 <3` |
| `brace-expansion@>=5.0.0 <5.0.7` | `>=5.0.7 <6` |

Why this avoids the break #854/#853 hit: `@eslint/config-array` →
`minimatch@3.1.5` → `brace-expansion@1.1.16` (the fixed **v1** line, unchanged
export shape). `5.0.7` is pulled only by `minimatch@10.2.5` consumers
(`@typescript-eslint/typescript-estree`, openclaw) that tolerate it. The
`expand is not a function` error came from dragging config-array's v1 across a
major to 5.x; the major-bounded overrides never do that. The `osv-scanner.toml`
ignore for GHSA-3jxr is removed.

**2. astro osv triage — hardened**

Added `ignoreUntil = 2026-10-21` to the three residual astro ignores so the
exception can't outlive #851 silently, and grounded each reason in the
advisories' real scope: all three are XSS bugs upstream scopes to on-demand SSR
/ the dev server only — Starlight emits static HTML, so the built-and-deployed
docs site is unaffected regardless of the deferred Astro 7 major.

## Not re-touched

Everything #853 already landed — astro 6.4.8, body-parser 2.3.0, js-yaml 4.3.0,
protobufjs 8.7.1, and the openclaw unused-ignore cleanup — is inherited from
main and does not appear in this diff.

## Verification

- `pnpm install --frozen-lockfile`: lockfile up to date
- brace-expansion resolves to 1.1.16 / 2.1.2 / 5.0.7; `@eslint/config-array`
  stays on the fixed v1 (1.1.16)
- `osv-scanner.toml` parses; prettier clean
- Full CI (lint / typecheck / test / osv-scan) runs on this branch

Closes #854


clemenshelm:
Superseded by main's consolidated OSV fix (`12accd496` + `40544662e`), which landed in parallel and is **more correct**: it discovered that bumping `brace-expansion` to 5.0.7 breaks `@eslint/config-array`, so it ignores that advisory (tracked in #854) instead of forcing the bump as this PR did. Main already covers all 12 advisories — same body-parser/js-yaml/protobufjs overrides, same docs astro→6.4.8, same 3 astro IgnoredVulns pointing at #851 (this PR's tracking issue, now in use). Closing as redundant; nothing here to salvage.

clemenshelm:
Superseded by #857 — same branch (`security/osv-transitive-bumps`), rebased onto main after #853 merged (couldn't be reopened once the history was rewritten). #857 carries the real brace-expansion fix (closes #854) plus the astro `ignoreUntil` hardening.

Issue #705, opened 2026-07-12
Conversational Inbox-Agent workflow creation from agent chat (propose, don't self-activate)


clemenshelm:
## Goal

Fast-follow to the **Inbox Agent (#139)**: let a user create an email workflow by describing it in natural language to an agent ("whenever an invoice arrives in the invoices mailbox, file it in Odoo with the attachment"), instead of only through the structured Automations form. The agent translates the request into the canonical structured workflow object.

## Design

- **New tool** (e.g. `create_inbox_workflow` in a `pinchy-automations` plugin) maps the NL description to the structured workflow fields (connections, filter, action prose, cadence) and writes it via an internal Pinchy route.
- **Propose, don't self-activate.** The tool creates a **pending / disabled** workflow, never an active one. A human reviews the *structured* translation in the Automations tab and enables it. Rationale: a workflow is standing autonomous authority; an agent must not grant itself that. Mirrors OpenClaw's own "skill-creator can't bypass workshop proposals" guard and our draft-only philosophy.
- **Same object, one system.** Both the form (#139) and the chat tool write the identical workflow schema — the tool is only an NL→fields translator. Uniform audit and tracking, no parallel system.
- **Audited on both layers:** `tool.create_inbox_workflow` (the dispatch) and the resulting `workflow.created`.

## Access (scope-based, not role-based)

- A member may create a workflow on a **personal agent + own connections**, self-confirmed via the tool-call confirmation gate (#517).
- A workflow touching a **shared agent or shared / org mailbox** requires an **admin**.

**Current reality (verified in code):** every user already has their own **personal** Smithers (`isPersonal: true, ownerId: <user>`, auto-seeded on invite claim via `seedPersonalAgent`; see `packages/web/src/lib/personal-agent.ts`). Additional *personal* agents can't be created, but **shared** agents can (templates, `isPersonal: false`, via `POST /api/agents`). So the member-self-service branch is **live from day one** — this is **not** admin-only: a user creates a workflow on their own Smithers + own connections (self-confirmed via #517); only workflows on shared agents / shared mailboxes need an admin. Notifications for a personal-Smithers workflow go to its owner.

## Dependencies

- **#517** (tool-call confirmation, #124 Tier 1+2) — the gate that makes the create action safe and inline; must land first.
- **#139** Inbox Agent — provides the workflow object + structured form this builds on.
[...]

Issue #139, opened 2026-04-15
Inbox Agent: one trigger → many item runs with draft-in-source-system pattern


clemenshelm:
## Problem

All three pilot customers (Gittermattenzaun, APSA, Portatour) converge on the same high-value pattern:

> **Incoming input (ticket / email / request) → Draft of a response/action in the source system → Human approves & sends.**

This is not a scheduled-briefing use case — it's an **item-processing** loop: one trigger fires repeatedly, produces many independent runs (one per new item), and the outcome lives in the external tool (Zendesk, Odoo Helpdesk, invoice inbox).

## Goal

**Inbox Agent** — a higher-level concept on top of OpenClaw Cron, where a single scheduled trigger fans out into **N independent runs**, one per new source item, with durable state-tracking so items are never processed twice.

## Scope (to be refined)

- Polling trigger (cron-based) that queries an external source for new items since last successful poll
- **Fan-out:** one trigger tick → one run per new item, with isolated sessions
- **State-tracking:** which items have been processed (per-agent)
- **Draft target:** artifacts land in the source system (Zendesk ticket reply, Odoo mail.message draft, etc.) — Pinchy stores only the summary and link
- **Activity feed:** one notification per processing batch ("3 new tickets processed, drafts created") with expandable details
- Integration examples for launch: Odoo Helpdesk, Zendesk, IMAP inbox

## Why this is Phase 2

Phase 1 (Scheduled Briefings, #138) builds the shared foundation:
- Activity feed + per-user notification fan-out
- Reconciliation loop for Pinchy↔OpenClaw consistency
- Usage dashboard integration
- Audit trail for background runs

Once those primitives are live, Inbox Agent is primarily about:
- Generalizing `briefing_runs` → `agent_runs` with trigger-source abstraction
- Adding the polling+state-tracking layer
- Per-source integrations (Odoo, Zendesk, IMAP)

## Customer Evidence

- **Gittermattenzaun (Odoo):** Support-Inbox → ticket triage + response drafts with Odoo context. Also: quote drafts from natural-language requests.
- **APSA (staffing):** Invoice workflow automation with strict approval gates. Billing/payment matching drafts.
- **Portatour:** Multilingual support-email drafting using Zendesk KB as context.

All three explicitly want **draft-only, no auto-send**.

## Acceptance Criteria (rough)

- [ ] Admin can configure an Inbox Agent: source system, polling interval, prompt template per item
- [ ] Items are processed exactly once (state-tracked)
[...]

PR #864, opened 2026-07-21
feat(automations): write-API to create Inbox Agent email workflows (#139)

clemenshelm:
## What

The **Automations write-API** — the single write path the Inbox Agent (#139) has been missing. On `main` today, `email_workflows` rows are inserted **only in tests**; there is no production create path, no shared schema, no `*.created` audit. This adds all three.

`POST /api/automations` maps one shared Zod schema (`lib/schemas/automations.ts`) to a transactional two-table write (`email_workflows` + `email_workflow_connections`). Both the Automations form (#139) and — after #517 — the conversational create tool (#705) write this exact shape, so every way a workflow is born shares **one contract, one RBAC gate, one audit event** ("same object, one system").

## Why this, and why now

#705 (conversational create) sits on top of a stack of unbuilt bricks: it needs #517 (tool-call confirmation, still open), a review UI, **and** this write-API. This is the unblocked foundation both #139's form and #705's tool depend on, and it's independent of #517.

## Design

- **Scope-based RBAC** (design §7, #705): a member may create a workflow on a **personal agent they own**; a **shared agent** requires an **admin**. "Own connections" maps to the connections the agent is permitted to read (`agent_connection_permissions`) — `integration_connections` has no per-user owner, so agent-scoped permission is the real, code-backed boundary. An unknown connection id has no permission row either, so one check rejects both "no access" and "no such connection".
- **Propose, don't self-activate:** always written `pending` + `disabled`. Each connection is attached with a **now-watermark** (`since_ts = now`) so a fresh workflow never retroactively processes historical mail (design §8). Enablement is a separate, human-gated step — deliberately **not** a request field.
- **Audit:** new `AuditResource` value `email_workflow` → `email_workflow.created` (covers `.updated`/`.deleted` for the coming UI lifecycle too). `deferAuditLog` (rows committed, non-rollbackable). **IDs only** in `detail` — connection names can carry addresses.

## Naming note

The issue sketch says `workflow.created`; this uses **`email_workflow.created`** to match the `email_workflows` table + `lib/email-workflows` exactly (`workflow` alone is ambiguous).

## Tests (TDD, RED-first)

- **Schema unit** (`automations.test.ts`): defaults, ≥1 connection, blank-name/agent/action rejection, sweep-window bounds.
- **Real-DB route integration** (`automations-create.integration.test.ts`): member-on-own-personal-agent → 201 pending+disabled with now-watermark + `email_workflow.created` payload; member-on-shared → 403; member-on-others-personal → 403; admin-on-shared → 201; unpermitted/unknown connection → 400 **with no partial write**; unknown agent → 404; unauthenticated → 401.

## Gates

`typecheck` ✅ · `eslint` ✅ (0 errors) · `prettier --check` ✅ · schema unit ✅ (6) · route integration ✅ (8) · full web unit suite: 9101 passed. One unrelated flake (`openclaw-config.test.ts` flush-disable resolver) fails only under full-suite load and **passes 216/216 in isolation** — not touched by this diff (audit union + new automations files; the unit suite doesn't load `*.integration.test.ts`).

## Docs

No user-facing docs yet: the Automations feature has no UI surface (that arrives with #139's form / the Automations tab). The design of record lives in `docs/plans/2026-07-12-inbox-agent-design.md`. Docs land with the UI PR.

## Out of scope / follow-ups

- Automations tab UI (list / review / enable-disable) — #139.
- Conversational create tool — #705 (needs #517 first).
- `email_workflow.updated` / `.deleted` routes — land with the management UI.

### DIFF
--- a/packages/web/src/__tests__/api/automations-create.integration.test.ts
+++ b/packages/web/src/__tests__/api/automations-create.integration.test.ts
@@ -1,0 +1,244 @@
+// Real-DB integration tests for POST /api/automations — the single write path
+// for the Inbox Agent's email workflows (design §5; the foundation both the
+// Automations form #139 and the conversational create tool #705 write through).
+//
+// Why real DB, not mocked chains: the load-bearing behavior here is a
+// transactional two-table write (email_workflows + email_workflow_connections)
+// with a per-connection watermark, plus scope-based RBAC that queries the
+// agent's ownership and its connection permissions. Mocking @/db would assert
+// nothing about any of that. So @/db runs for real against a freshly migrated
+// Postgres (global-setup.ts), truncated between cases (setup.ts).
+//
+// What stays mocked, and why:
+//   - @/lib/auth.getSession — the only way to drive the withAuth scope branches
+//     (member vs admin, owner vs not) deterministically.
+//   - @/lib/audit-deferred.deferAuditLog — a thin next/after wrapper whose
+//     callback does not run under a direct handler call; we assert the payload
+//     it was handed instead of round-tripping through `after()`.
+import { describe, it, expect, vi, beforeEach } from "vitest";
+import { eq } from "drizzle-orm";
+
+import { db } from "@/db";
+import {
+  agents,
+  users,
+  emailWorkflows,
+  emailWorkflowConnections,
+  agentConnectionPermissions,
+  integrationConnections,
+} from "@/db/schema";
+import { makeNextRequest, routeContext } from "@/test-helpers/route";
+
+const { getSessionMock, deferAuditLogMock } = vi.hoisted(() => ({
+  getSessionMock: vi.fn(),
+  deferAuditLogMock: vi.fn(),
+}));
+
+vi.mock("next/headers", () => ({
+  headers: vi.fn().mockResolvedValue(new Headers()),
+}));
+
+vi.mock("@/lib/auth", () => ({
+  getSession: getSessionMock,
+  auth: { api: { getSession: getSessionMock } },
+}));
+
+vi.mock("@/lib/audit-deferred", () => ({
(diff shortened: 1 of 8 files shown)


clemenshelm:
Review follow-up pushed as abe4dba7f — five hardening fixes on top of the original diff:

1. **Audit `resource` pointer**: the `email_workflow.created` entry now carries `resource: email_workflow:<id>`, matching every other `*.created` event so the row correlates in resource-based audit queries.
2. **PII scrub**: the free-text workflow name is passed through `scrubEmails()` before landing in the append-only, HMAC-signed audit detail (same treatment as the IMAP route's `connectionName`). The stored row and the API response keep the raw name.
3. **Read-capable gate**: the connection check now requires a READ grant (`EMAIL_READ_OPERATIONS` = `read` + legacy `search`/`list` aliases, newly exported from `tool-registry`) instead of any email-model row — a draft/send-only grant can no longer anchor a workflow whose trigger must list and read mail. Covered by two new integration tests (send-only → 400, legacy `search` → 201).
4. **Curated union**: `email_workflow.created` added to `AuditEventType` (the `${AuditResource}.created` template satisfied the write side invisibly), pinned by a type test.
5. **`connectionIds` cap**: `AUTOMATION_MAX_CONNECTIONS` (50) bounds the 400-error echo and the audit `detail.connectionIds` list against the 2048-byte detail cap.

Also moved the schema unit test to `src/__tests__/lib/schemas/` where every other schema test lives (net-zero move).

Gates: typecheck ✅ · eslint 0 errors ✅ · prettier ✅ · full unit suite 9102 passed (the two failures — `openclaw-config` flush-disable and `timezone-settings` save — are the known full-suite-load flakes, both pass 227/227 in isolation, neither file touched here) · full `test:db` integration suite 319 passed against a fresh `pgvector/pgvector:pg17-trixie` Postgres.

PR #915, opened 2026-07-24
fix(chat): follow tab visibility for WS reconnect lifecycle (#895)

### LINKED ISSUE #895: Chat WebSocket lifecycle should follow tab visibility (pause reconnect when hidden, close after grace, instant resume on visible)
## Summary
`packages/web/src/hooks/use-ws-runtime.ts` reconnects with backoff but gives up after `MAX_RECONNECT_ATTEMPTS = 10` (~45 s, `reconnectExhausted`) and has no visibility handling. Two consequences today:

1. **Dead-end after laptop sleep / long background:** by the time the user returns to the tab, the 10 attempts are long spent — red connection indicator, only a full reload recovers. Affects every deployment.
2. A hidden tab keeps retrying (while attempts last) and keeps its WS open indefinitely, holding server resources for nobody.

## Request
Make the connection lifecycle follow the Page Visibility API:

- **Tab hidden →** pause reconnect attempts (no dialing from background tabs).
- **Tab hidden longer than a grace period (~5 min) →** close the WebSocket cleanly.
- **Tab visible →** reset the reconnect counter and reconnect immediately (also from the `reconnectExhausted` state).

Net effect: an open WS means someone is actually looking at the tab; a returning user reconnects instantly without reload; background tabs stop burning retries and connections.

## Notes
- Generic UX/resource improvement; also a prerequisite for hosted trial-instance sleep/wake (an instance can be stopped without its abandoned tabs re-waking it, and a returning visible tab wakes it automatically through its reconnect attempts).
- Reconnect must tolerate a ~10–20 s unavailable window (e.g. server restart): visible-tab retries should keep going at the capped 5 s interval rather than exhausting quickly.
- Existing reconnect/history-reconciliation semantics (server history frame as source of truth) must be preserved; see the reconnect paths in `use-ws-runtime.ts`.

clemenshelm:
## Summary

Fixes #895. `use-ws-runtime.ts` reconnects with exponential backoff but gave up after `MAX_RECONNECT_ATTEMPTS = 10` (~45s), landing in `reconnectExhausted` with no way back short of a full reload — the common case after laptop sleep or a long background tab. There was also no way to pause reconnect attempts or free a stale connection while the tab is hidden.

Makes the WebSocket lifecycle follow the Page Visibility API:

- **Hidden** → pause reconnect attempts. If a backoff timer is already scheduled when the tab goes hidden, it fires but does not dial (dropped, not rescheduled) — no attempts are burned in the background.
- **Hidden longer than a 5-minute grace period** (`HIDDEN_CLOSE_GRACE_MS`) → close the WebSocket cleanly. Reuses the existing `suspendForPageLifecycle`/`recoverFromPageLifecycle` lifecycle-suspend path (already used for `pagehide`/`freeze`/`offline`), so the close does not itself schedule a reconnect. Shorter hidden spells (< 5 min) leave the connection open — no wasted reconnect+history-reload round trip for a quick tab switch.
- **Visible** → reset the reconnect counter and reconnect immediately, including recovering out of the `reconnectExhausted` dead end. Once reconnecting, retries continue at the existing capped 5s interval, tolerating a ~10-20s server-unavailable window — this is also a prerequisite for hosted trial-instance sleep/wake, where a returning visible tab should wake a stopped instance through its own reconnect attempts rather than needing external help.

Existing reconnect/history-reconciliation semantics (server history frame as source of truth after reconnect, `pagehide`/`freeze`/`offline` immediate-suspend paths) are unchanged.

## Changes

- `packages/web/src/hooks/use-ws-runtime.ts`:
  - New `HIDDEN_CLOSE_GRACE_MS` constant (5 min) and `hiddenGraceTimerRef`.
  - `handleVisibilityChange`'s hidden branch now schedules a grace-period close (`scheduleHiddenGraceClose`) instead of closing immediately.
  - New `handleTabVisible`: cancels the grace timer, delegates to `recoverFromPageLifecycle` if the connection was deliberately suspended, otherwise resets the reconnect counter/`reconnectExhausted` and reconnects if not already open/connecting.
  - The reconnect backoff timer now checks `isTabHidden()` before dialing and simply drops the deferred attempt if hidden.
  - `hiddenGraceTimerRef` cleared on effect cleanup (agent/chat switch and unmount) — no leaked timers.

## Tests

New file `packages/web/src/__tests__/hooks/use-ws-runtime-visibility.test.tsx` (12 tests):
- `does not dial a pending reconnect while hidden, and does not consume further attempts`
- `keeps the WebSocket open when hidden for less than the grace period`
- `closes the WebSocket cleanly after the hidden grace period elapses, with no reconnect scheduled`
- `reconnects immediately with a reset counter when the tab becomes visible after reconnectExhausted`
- `keeps retrying at the capped 5s interval through a simulated ~20s outage after becoming visible, without premature exhaustion`
- `clears the hidden-grace timer on unmount, leaving no leaked timers`

Red→green: verified by temporarily reverting the implementation change (`git stash`) and running the new suite — 3/6 failed against the pre-existing code (the grace-period-open, exhaustion-recovery, and grace-timer-cleanup cases, i.e. exactly the genuinely new behavior; the other 3 passed incidentally under the old immediate-close-on-hidden behavior). Restoring the implementation makes all 6 green.

Existing `use-ws-runtime.test.ts` suite (including its pre-existing "resumes from sleep" visibilitychange test) passes unmodified.

## Test plan

- [x] `pnpm -C packages/web test` — 9379 passed, 16 skipped (694/695 files passed, 4 pre-existing skips unrelated to this change)
- [x] `pnpm -C packages/web typecheck` — clean
- [x] `pnpm -C packages/web lint` — 0 errors (pre-existing warning count unchanged)
- [x] `pnpm format:check` (repo root) — clean
- [x] `pnpm test:scripts` (repo root) — 379 passed

### DIFF
--- a/packages/web/src/__tests__/hooks/use-ws-runtime-visibility.test.tsx
+++ b/packages/web/src/__tests__/hooks/use-ws-runtime-visibility.test.tsx
@@ -238,6 +238,43 @@ describe("useWsRuntime — visibility-driven reconnect lifecycle (#895)", () => 
     expect(historyRequestCount()).toBe(2);
   });
 
+  it("arms the #956 watchdog on the visible reconcile, so a stalled answer still self-heals", () => {
+    // The visible-with-OPEN-socket reconcile is a history request the UI
+    // waits on, so it must carry the same deadline as every other one
+    // (sendHistoryRequest's invariant). The socket that "survived" the
+    // hidden spell is exactly the one that may be half-dead — still
+    // reporting OPEN after a network change while the tab wasn't looking,
+    // with every send going nowhere (the forceReconnect scenario). Without
+    // the watchdog, pendingHistoryRef then sticks forever: focus- and
+    // poke-driven catch-ups both coalesce on it and skip, no timeout is
+    // surfaced, and the chat hangs until a full reload.
+    const { result } = renderHook(() => useWsRuntime("agent-1"));
+    const ws1 = latestWs();
+
+    act(() => {
+      ws1.simulateOpen();
+    });
+
+    act(() => {
+      setVisibility("hidden");
+    });
+    act(() => {
+      setVisibility("visible");
+    });
+    // Reconcile request went out over the still-open socket; nobody answers.
+    expect(ws1.close).not.toHaveBeenCalled();
+
+    // Advance past the 20s history-answer deadline while visible: the
+    // watchdog must surface the stall and self-heal via forceReconnect
+    // (deliberate close of the half-dead socket).
+    act(() => {
+      vi.advanceTimersByTime(20_000);
+    });
+
+    expect(result.current.historyTimedOut).toBe(true);
+    expect(ws1.close).toHaveBeenCalledTimes(1);
+  });
+
   it("closes the WebSocket cleanly after the hidden grace period elapses, with no reconnect scheduled", () => {
     const { result } = renderHook(() => useWsRuntime("agent-1"));
     const ws1 = latestWs();
@@ -437,6 +474,13 @@ describe("useWsRuntime — visibility-driven reconnect lifecycle (#895)", () => 
 
     act(() => {
       setVisibility("visible");
+    });
+
+    // Answer the visible-reconcile history request so its (correctly armed)
+    // #956 watchdog disarms — this test pins only the grace-timer cancel,
+    // not a stalled reconcile.
+    act(() => {
+      ws1.simulateMessage({ type: "history", messages: [], sessionKnown: true });
     });
 
     // The grace timer must be cancelled — advancing well past where it would
--- a/packages/web/src/__tests__/hooks/use-ws-runtime-visibility.test.tsx
+++ b/packages/web/src/__tests__/hooks/use-ws-runtime-visibility.test.tsx
@@ -168,6 +168,42 @@ describe("useWsRuntime — visibility-driven reconnect lifecycle (#895)", () => 
     expect(wsInstances).toHaveLength(1);
   });
 
+  it("re-requests history over an already-open socket on visible, even without an intervening close (#895 follow-up)", () => {
+    // e2e/18-tab-refocus-shrink.spec.ts pins the real-browser invariant this
+    // guards: a refocus must reconcile against canonical server history even
+    // when the grace period never elapsed and the socket never dropped —
+    // server-side state (e.g. an in-flight run whose history has since
+    // shrunk) can change while the tab wasn't listening. Before the grace
+    // period existed, EVERY hidden→visible cycle tore the socket down and
+    // reopened it, so a fresh history request was a side effect of the
+    // reconnect. The grace period made a short hide/show a no-op WS-wise —
+    // this test pins that the history re-request still happens even though
+    // the socket itself does not.
+    renderHook(() => useWsRuntime("agent-1"));
+    const ws1 = latestWs();
+
+    act(() => {
+      ws1.simulateOpen();
+    });
+
+    const historyRequestCount = () =>
+      ws1.send.mock.calls.filter((call: string[]) => JSON.parse(call[0]).type === "history").length;
+    expect(historyRequestCount()).toBe(1);
+
+    act(() => {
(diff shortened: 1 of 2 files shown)


clemenshelm:
Addressed spec review feedback in 0948d090: \`handleTabVisible\`'s \`lifecycleSuspendedRef\` branch returned before resetting \`reconnectAttemptRef\`/\`reconnectExhausted\` — \`recoverFromPageLifecycle()\` never touches that counter itself. Composing behaviors 2+3 (exhaust while visible → hide past the 5-min grace close → return while the server is still down) replayed exactly one stale attempt and flipped straight back to \`reconnectExhausted\` instead of restarting the full capped-backoff budget.

Fix: hoisted the reset above every branch in \`handleTabVisible\`, including the already-OPEN no-op (harmless) and the CONNECTING early return (an in-flight dial that fails now also gets a fresh budget).

New/updated tests:
- \`restarts the full backoff budget on return from a grace-period close, even if the counter was already exhausted before going hidden\` — red against the pre-fix code (reproduced the exact reviewer-described bug), green after the hoist.
- \`visible just before the grace period elapses cancels the pending grace close\` — pins the already-correct grace-cancel behavior per the reviewer's cheap-test note.

Gates: targeted hooks suite 335/335 passed; full \`pnpm -C packages/web test\` 9381 passed, 16 skipped (0 failures — an intermittent, pre-existing jsdom rAF/timer teardown flake in the unrelated \`runtime-stability.test.tsx\` file surfaced in 2 of 4 full-suite runs during verification, reproducing with and without this diff and not in isolation; not a regression from this change); typecheck clean; lint 0 errors (pre-existing warning count unchanged); \`pnpm format:check\` clean.

clemenshelm:
Two more rounds pushed (0948d090..5115bbaf):

**E2E fix** (\`e2e/18-tab-refocus-shrink.spec.ts\` failure): confirmed the hypothesis — before the grace period, every hidden→visible cycle closed+reopened the WS, so a refocus always re-requested history as a side effect of the reconnect. With the grace period, a short hidden spell leaves the socket OPEN and \`handleTabVisible\`'s OPEN branch was a no-op — no history re-request, silently breaking the refocus-reconcile invariant the E2E pins. Fixed: the OPEN branch now sends a fresh history request over the live socket (same buffer-then-drain protocol as every other history-on-open-socket path).

**Rebase onto origin/main** (180 commits; only one touched this file — d75afbb9, the #956 history-answer deadline/watchdog): resolved the one real conflict (constants block, both intents kept), then integrated properly — routed my new OPEN-branch history request through the new \`sendHistoryRequest\` helper so it's covered by the watchdog like every other path. That surfaced a genuine bug: the watchdog previously stood down only via the OLD immediate-suspend-on-hidden behavior; with the grace period, going hidden no longer suspends immediately, so an outstanding request's watchdog could fire mid-hidden-period and force its own reconnect, bypassing the hidden-aware pause entirely (and in the canonical suite this cascaded into a cross-test \`document.visibilityState\` leak once the resulting assertion failure skipped a test's own cleanup line). Fixed by clearing the watchdog when scheduling the grace close too; it's freshly re-armed by whichever path brings the tab back.

New unit tests (TDD, red→green against a temporary revert each time):
- \`re-requests history over an already-open socket on visible, even without an intervening close\` (E2E fix)
- \`does not let the #956 history-answer watchdog fire while hidden, even past its deadline\` (rebase-integration fix)

Local E2E reproduction: not run — requires a dedicated Postgres test stack (port 5433 + migrations + full dev server) not present in this worktree; relying on CI's \`E2E Tests\` job, now queued fresh on the rebased branch.

Gates after rebase + both fixes: targeted hooks suite 350/350 passed; full \`pnpm -C packages/web test\` 10337 passed, 20 skipped, 0 failed; typecheck clean; lint 0 errors (10 pre-existing warnings, confirmed identical on origin/main); \`pnpm format:check\` clean; \`pnpm test:scripts\` 551 passed.

osv-scan: per the note, main already carries the postcss/tar fixes — the rebase should resolve that failure too.

PR #920, opened 2026-07-25
fix(deps): bump postcss, tar, brace-expansion past fresh osv advisories

clemenshelm:
## What & why

`Security audit / osv-scan` (`fail-on-vuln: true`) is **red on `main`** — three fresh advisories landed against deps already present in the lockfiles, so the scan blocks **every** open PR (e.g. #919), not any one change. This is the documented "fresh advisory reddens main" pattern: the fix is a dep bump, in its own focused PR so it unblocks all of them at once.

## Advisories (6 pkgs / 7 vulns, all High/Medium)

| Package | in lock | → | Advisory | Disposition |
|---|---|---|---|---|
| `postcss` | 8.5.10 (docs), 8.5.16 (root) | **8.5.23** | GHSA-6g55-p6wh-862q, GHSA-r28c-9q8g-f849 | **fixed** |
| `tar` | 7.5.19 | **7.5.22** | GHSA-r292-9mhp-454m (stack-overflow DoS) | **fixed** |
| `brace-expansion` (5.x) | 5.0.7 | **5.0.8** | GHSA-mh99-v99m-4gvg (OOM DoS) | **fixed** |
| `brace-expansion` (1.x/2.x) | 1.1.16, 2.1.2 | *(kept)* | GHSA-mh99-v99m-4gvg | **filtered — unfixable, unreachable** |

## The fix — override bumps only (no source change)

The repo already pins security minimums via `pnpm.overrides`; these had been overtaken:

- **postcss** `<8.5.10 → >=8.5.18`, in **both** root and `docs/package.json` (`docs` is a standalone workspace with its own lockfile).
- **tar** `<7.5.16 → <7.5.21 : >=7.5.21`.
- **brace-expansion** — the **only** patched version is `5.0.8` and the advisory has **no per-major backport**. I bump the **5.x** override to it (`<5.0.8 → >=5.0.8 <6`, used by `minimatch@10`). The **1.x/2.x** lines stay at their latest-in-major: forcing them to 5.0.8 **breaks `minimatch@3.1.5`** (still pulled under `@eslint/config-array`), which expects brace-expansion's v1 API and crashes `eslint` with `TypeError: expand is not a function` (exit 2). Those two lines are build/lint/test tooling matching file paths against **developer-authored** globs — no untrusted input, in no runtime image — so the OOM DoS is not reachable. Filtered in `osv-scanner.toml` (`ignoreUntil = 2026-10-21`), **tracked in #921**, cleared once the `minimatch@3.x` consumers are gone.

`nanoid` 3.3.11/3.3.15→3.3.16 is postcss's own patch-level transitive. **No other package versions change.**

## Verification

- `osv-scanner --config=./osv-scanner.toml` over **both** lockfiles → **"No issues found"**.
- `pnpm -C packages/web lint` → **0 errors** (the per-major brace-expansion split keeps `minimatch@3.1.5` working — this is why the naïve single-override collapse was reverted).
- `pnpm install --frozen-lockfile` succeeds → `package.json` + lockfiles frozen-consistent.
- No app code touched; `docs/pnpm-lock.yaml` counts as **code** for CI path-filtering (`vuln-scan` reads it), so the security jobs run on this PR.


clemenshelm:
Superseded — `main` already carries this fix.

While this PR was open, the same three advisories were handled directly on `main`:

- `8b39d074b` postcss `>=8.5.18` + tar `>=7.5.21` overrides
- `b46eb9711` → `bcf06586a` brace-expansion: first forced `>=5.0.8` tree-wide, then walked back to **patch the 5.x line only + ignore the unfixable 1.x/2.x** — the identical conclusion this PR reached, after hitting the identical `minimatch@3.1.5` breakage.

I verified `main`'s resolved state matches this branch: root lockfile postcss **8.5.23**, tar **7.5.22**, brace-expansion **1.1.16 / 2.1.2 / 5.0.8**; docs postcss **8.5.22** (≥8.5.18, so GHSA-r28c-9q8g-f849 is satisfied). `main`'s `osv-scanner.toml` reason is also **more complete** than mine — it identifies `archiver@5` (via exceljs/pinchy-files) as a second `minimatch@3` consumer, which I had missed.

Closing rather than force-landing a duplicate. Nothing here is lost.

Issue #939, opened 2026-07-28
kb: serve the converted PDF as the preview for an Office source


clemenshelm:
**Wave 2.** Depends on #934 (download affordance) and #936 (artifact store).

## Scope

A cited Office document opens in the same viewer as a PDF, showing the **converted artifact**, at the cited spot. The download control then offers **both**: the original (the file on their drive, the one to send a customer) and the converted PDF (sometimes exactly what they want to send anyway — it saves them the conversion).

`workspace-file` currently resolves a request to a file under the agent's allowed paths. It needs to resolve an Office path to its artifact for the *preview* while continuing to serve the *original* for download — without widening what a request can reach: the artifact store is outside `/data`, so path containment has to be re-reasoned rather than reused. This is a file-exfiltration surface; see the containment notes in `agent-file-access.ts`.

## Owns

`workspace-file/route.ts`, `attachment-preview.tsx`, `source-links.ts` (extension allow-list widening), their tests.

## Acceptance

- [ ] Office citation opens a rendered preview at the cited spot
- [ ] Both downloads offered, correctly named and distinguishable
- [ ] Requesting an artifact for a document outside the agent's allowed paths is refused — with a test
- [ ] A stale artifact (original changed since conversion) is not served
- [ ] `next.config.ts` frame-options coverage still passes — `frame-options-route-coverage.test.ts` fails CI if a serving route lacks its entry

Issue #938, opened 2026-07-28
kb: index Office documents through the converted PDF


clemenshelm:
**Wave 2.** Depends on #933 (locator model) and #936 (conversion + artifact store). Both must land first; this issue is the integration and owns `ingest.ts`, which is why Wave 1 deliberately stayed out of it.

## Scope

- Extend `DEFAULT_ALLOWED_EXTENSIONS` (`lib/knowledge/exclude-globs.ts`, today `[".pdf"]`) with the page-shaped Office formats.
- During ingest: convert via #936, index the **converted PDF's** text.
- **Identity stays the original.** The document a citation names is `report.doc`, the file on the user's drive. The converted PDF is a preview artifact and must never appear as the cited document — a path the reader cannot find on their share destroys the citation.
- Emit the right locator per format (#933): **heading path** for Word, **slide number** for PowerPoint. Not "page" — Word pagination is a rendering result and would be false of the original.
- Word without headings: emit no locator. The link still navigates to the right spot in the preview; an omitted locator is better than one that does not match what the reader sees in Word.
- Gate indexing on #936's verification chain. A document that fails it is **not indexed** and therefore appears in the unreadable list (#935) automatically — that list derives from "row exists, no chunks", so nothing needs wiring.
- Respect the exit-0 / exit-137 split from #936: a memory failure must not mark a document unreadable.

## Hidden content

Conversion can surface material the author hid — tracked changes, comments, speaker notes. That is not only a leak; it makes the citation **unverifiable**, because the reader opens the original and cannot find the sentence. Decide explicitly and test the decision.

## Owns

`lib/knowledge/ingest.ts`, `lib/knowledge/exclude-globs.ts`, their tests, the ingest section of the KB guide.

## Acceptance

- [ ] `.doc/.docx/.ppt/.pptx` indexed; citation names the **original** path, never the artifact
- [ ] Word emits heading locators, PowerPoint slide numbers; a test asserts Word never emits a page locator
- [ ] Idempotency preserved: unchanged file skipped, changed file re-converted and re-indexed, deleted file dropped
- [ ] A conversion failure leaves the document out of the index and visible in #935
- [ ] Test with old data: PDFs indexed before this change keep working untouched (AGENTS.md § *Test Migrations Against Pre-Existing Data*)

clemenshelm:
## Measured: the Word heading path does NOT fall out of the converted PDF

[...]

Issue #937, opened 2026-07-28
kb: read spreadsheets directly, anchored on sheet and row range


clemenshelm:
Part of the Office/scanned-document work. **Wave 1 — parallel with the other Wave-1 issues.** Deliver a standalone, tested extractor; wiring into the indexer is a separate Wave-2 issue.

## Why spreadsheets do not go through PDF

The sibling issue converts page-shaped Office formats to PDF. Spreadsheets deliberately do not, for three independent reasons:

**The anchor would be meaningless.** A 40-column sheet rendered to PDF becomes an unreadable mess spread over dozens of pages, and a citation reading *"page 17"* gives the reader nothing they can verify. The intrinsic anchor is **sheet + row range**, and unlike a Word page it is stable — it is a property of the document, not of a renderer.

**Page-fitting is the step that silently clips.** Text that does not fit a column is cut when the sheet is laid out onto pages. That is the exact failure the conversion tripwire exists to catch — and reading cells skips the step entirely, so it cannot happen. Reading cells is *higher* fidelity here, not lower.

**It saves 422 → the calc packages.** Measured: `libreoffice-writer libreoffice-impress fonts-liberation` costs +422 MB over `node:22-slim`; the full `libreoffice` costs +525 MB. Not needing `libreoffice-calc` is worth ~103 MB.

Nobody sends a customer a thirty-page Excel PDF anyway, so the "they may want to forward the PDF" argument does not apply to this format.

## Prior art to adopt rather than invent

Current practice for tabular RAG is **structure-aware chunking**: each row as a key-value block inside a row-tree hierarchy, with token-constrained splitting and merging, rather than fixed-size text chunks. Fixed-size chunking is documented to tear apart row-level structure and cross-sheet relationships. Follow that shape; the sheet+row-range anchor falls out of it naturally.

References:
- Structure-Aware Chunking for Tabular Data in RAG — https://arxiv.org/pdf/2605.00318
- Sheet as Token: multi-sheet spreadsheet understanding — https://arxiv.org/pdf/2605.05811

## Dependency

There is **no xlsx library in the repo today** (checked). Add `exceljs` (MIT — AGPL-compatible). SheetJS community edition (Apache-2.0) is an acceptable alternative if exceljs turns out not to read some real file; state the reason in the PR if you switch.

## Scope

- Read `.xlsx` (and `.xls` if the chosen library handles it — verify, do not assume) into rows with their sheet name and row numbers.
[...]

PR #958, opened 2026-07-28
feat(kb): convert page-shaped Office documents to PDF, with a verification chain

### LINKED ISSUE #936: kb: convert page-shaped Office documents to PDF, with a verification chain
Part of the Office/scanned-document work. **Wave 1 — parallel with the other Wave-1 issues.** Deliver a standalone, tested function plus its artifact store; wiring into the indexer is a separate Wave-2 issue, and that separation is what keeps this out of `ingest.ts` where two other issues would collide with it.

## Why convert rather than extract natively

Office support is a product requirement across customers, **legacy formats included**. In the Noack sample corpus (222 files) the Office share is 19 files — and **13 of them are legacy binary** (`.doc` 11, `.ppt` 2):

| | count |
|---|---|
| `.doc` | 11 |
| `.docx` | 4 |
| `.xlsx` | 2 |
| `.ppt` | 2 |

The whole modern parsing ecosystem (Docling, MarkItDown, Unstructured, and anything built on `python-docx`/`python-pptx`) is OOXML-only and cannot read those 13. LibreOffice is the standard answer for legacy, and here it is not the lazy choice but the only one that covers the corpus.

Conversion also buys the thing the citation chain depends on: a **renderable artifact whose page numbers agree with what the reader is shown**. And it couples index and preview to one artifact, so a citation can never point at something the user cannot see.

Spreadsheets take a different path — see the sibling issue. `libreoffice-calc` is therefore **deliberately not installed**.

## Measured, so nobody re-derives it

Base image `node:22-slim` = 247 MB.

| variant | size | delta |
|---|---|---|
| `libreoffice-writer libreoffice-impress fonts-liberation` | 669 MB | **+422 MB** |
| full `libreoffice` | 772 MB | +525 MB |

Conversion of the 19 real corpus files: **17 converted, 15 s total**, slowest single file 1.46 s.

**Batch, do not spawn per file.** One process for 17 files: **14.2 s vs 25.6 s** one-per-file — process startup dominates.

**Memory is transient; LibreOffice is not a daemon.** Container floor, found by constraining:

| `--memory` | result |
|---|---|
| 128 MB | 0/17 — OOM |
| 160 MB | 11/11 `.doc` |
| **192 MB** | **17/17**, incl. the 5.9 MB `.ppt` |

(`memory.peak` under cgroup v2 reads 297–363 MB, but that includes reclaimable page cache — do not size from it.)

## Two failure modes, and they need different handling

**The exit code lies.** An input LibreOffice cannot load:

```
Error: source file could not be loaded     ← stderr only
Exit-Code: 0
Erzeugte Dateien: 0
```

A pipeline that trusts the return value books the document as converted and indexes nothing.

**OOM is distinguishable**: exit **137**. That difference is load-bearing:

- exit 0 + no artifact → **document-level failure**, final, belongs on the unreadable list
- exit **137** → **infrastructure too small**, retryable, must **not** mark the document unreadable

Without the split, one memory squeeze brands perfectly good documents as permanently unreadable and nothing ever retries them.

## Verification: a lower bound, not an equality check

Every conversion of the real corpus was measured against an independent extractor. **No case lost content.** All deviations were the *oracle* under-reporting:

- one `.docx` came out +25 % because it carries three headers and three footers (one 16 KB) that a raw `word/document.xml` read structurally cannot see
- `catppt` stopped after the title slide (39 words vs 868) **and mangled the characters**: `GODI? TORKAR`, `Dom?ale` where the converted PDF has the correct `GODIČ TORKAR`, `Domžale` — worse precisely on the Central/Eastern European diacritics the customer cares about
- `PF Aqua - Beverage Company Discussion Guide` exists as both `.doc` and `.docx`; both convert to **exactly 1511 words**

So an equality round-trip is **unbuildable** with these tools — it would fire constantly. Use a **lower bound**: `pdf_words >= orig_words × (1 − ε)`. In that direction a weak oracle can only miss a real loss, never raise a false alarm. It is a tripwire against the failure that actually hurts (clipped text), not a guarantee — do not describe it as one.

**Untested and worth knowing:** no document in the corpus has an overflowing fixed-size text box, so the clipping case the tripwire exists for has never actually fired. A synthetic fixture would be welcome.

## Artifact store

Converted PDFs **must not be written next to the original**: `/data` is mounted read-only and the docs state it as a product promise. Precedent to follow rather than reinvent: the `pinchy-pdf-cache` volume (`/var/cache/pinchy-files`) and `PdfCache`'s invalidation key — path + size + mtime + content hash + **`format_version`**. `format_version` is the one that matters: improving the converter must invalidate everything.

## Owns

- `Dockerfile.pinchy`, `docker-compose.yml` (artifact volume)
- new `packages/web/src/lib/knowledge/office-convert.ts` (+ artifact store) and tests
- `docs/` note on `PINCHY_MEM_LIMIT`

**Do not touch:** `lib/knowledge/ingest.ts`, the schema, `source-links.ts`, the `pinchy-knowledge` plugin, `attachment-preview.tsx`.

## Acceptance

- [ ] `.doc/.docx/.ppt/.pptx` → PDF; batch invocation, not one process per file
- [ ] Test reproducing **exit 0 + no artifact** and asserting it is reported as a document-level failure
- [ ] Test distinguishing **exit 137** and asserting it is *not* reported as document-level
- [ ] Lower-bound word check with the ε documented and justified
- [ ] Artifact store keyed incl. `format_version`; a bumped version invalidates
- [ ] Nothing is ever written under `/data`
- [ ] **Measure peak RSS during a real reindex that includes Office files** and document a `PINCHY_MEM_LIMIT` recommendation. The default is 1 GB and the same container runs the local embedding model during exactly that operation — that overlap has *not* been measured, and it should not be assumed to fit.

clemenshelm:
Closes #936. **Wave 1** — a standalone converter plus its artifact store. Wiring it into the indexer is #938 and serving the artifact as a preview is #939, which is why nothing here touches `ingest.ts`, the schema, `source-links.ts`, the `pinchy-knowledge` plugin or `attachment-preview.tsx`.

## What's here

- `lib/knowledge/office-convert.ts` — batch `.doc/.docx/.ppt/.pptx` → PDF, the failure split, the verification chain.
- `lib/knowledge/office-artifacts.ts` — the artifact store, keyed on `format_version` + content hash, on its own volume.
- `Dockerfile.pinchy` (+ `.dev`), `docker-compose.yml`, `entrypoint.sh` — LibreOffice and the `pinchy-kb-artifacts` volume.
- A `PINCHY_MEM_LIMIT` note in the deployment guide, backed by the measurement below.

## Three behaviours, each verified against the real binary rather than assumed

I probed `libreoffice-writer` in a container before writing the code, because two of these are exactly the kind of thing a fake would happily confirm wrongly.

**Batch, with index-named staging symlinks.** LibreOffice derives the output name from the input's basename, so `sub-a/Rechnung.doc` and `sub-b/Rechnung.doc` — the normal state of a real corpus — would write the same `Rechnung.pdf` and one document would end up serving the other's text. Staging as `0.doc`, `1.docx`, … makes the input→output mapping positional. Confirmed in the container: the output follows the *symlink's* name, not the target's.

**The exit code lies.** A truncated `.docx` prints `Error: source file could not be loaded` on stderr, exits **0**, writes nothing — and the other four files in the same batch convert normally. So the artifact's existence is the evidence, never the return value. Reproduced end-to-end in `office-convert.libreoffice.test.ts`.

**Exit 137 / SIGKILL is infrastructure, not a document verdict.** Both forms are handled (137 is what a shell or `docker` reports; a direct spawn sees the signal with a null code). A batch that was killed returns `infrastructure` for every file in it, so a memory squeeze can never brand good documents as permanently unreadable.

## Verification

`pdf_words >= source_words * (1 - 0.1)` against an *independent* extractor — mammoth for `.docx`, `catdoc`/`catppt` for the legacy binaries. Never equality: every measured deviation on the reference corpus was the oracle under-reporting, so an equality round-trip would fire constantly. Formats with no independent extractor (`.pptx`) are recorded `unverified`, deliberately not `ok`, and an oracle that breaks never fails a document.

The clipped-text-box case the tripwire exists for has never fired on real data, so the fixture for it is synthetic — that gap is documented in the code rather than papered over.

## Memory measurement

The acceptance criterion asks for peak RSS "during a real reindex that includes Office files". That reindex does not exist yet — Office extensions reach `ingest.ts` in #938 — so I measured the next best thing, and it is the part the issue flagged as unmeasured: **17 Word documents converting while the embedding model was actively embedding in the same container**.

| `PINCHY_MEM_LIMIT` | Result |
|---|---|
| `2g` | converts, fastest |
| `1g` (default) | 17/17 |
| `768m` | 17/17, ~20 % slower |
| `640m` | 17/17, ~70 % slower |
| `512m` | **exit 137, OOM-killed** |

So the default holds and the documented 4 GB VPS tuning (`768m`) holds; `512m` is below the floor. Sizing from `memory.peak` remains a trap — it tracks the ceiling because the mmapped GGUF and page cache are reclaimable.

Image cost is as the issue measured: `libreoffice-writer libreoffice-impress fonts-liberation catdoc` = **670 MB** (base `node:22-slim` is 247 MB). `libreoffice-calc` is deliberately absent — spreadsheets take a different path (#937/#940).

## Tests

38 tests across the three files, all passing in a Linux container **with real LibreOffice** — including a round trip that asserts `GODIČ TORKAR`, `Domžale` and `Maßnahmen` survive into the pdfjs text layer, which is what makes `fonts-liberation` non-optional.

`office-convert.libreoffice.test.ts` is gated with `describe.skipIf(!soffice)` (an allowed env/OS conditional gate, not an untracked skip) and skips on a developer laptop; the header documents the one-line `docker run` that executes it.

## Notes for the reviewer

- The docs section describes Office indexing, which becomes user-visible with #938. If #938 slips past a release, that section should be held back with it.
- Three tests fail in my local worktree and I checked them out on the base commit `48493b2c4` — they fail there too, so they are not from this change: `boot-inits.test.ts` (call-order expects 3, gets 6), `vitest-exclude-node-modules.test.ts`, and a flaky `pinchy-files` vision test. CI's `Lint, Test & Build` is green, so this looks environment-specific to that checkout rather than anything wrong on `main`.

### DIFF
--- a/Dockerfile.pinchy
+++ b/Dockerfile.pinchy
@@ -132,6 +132,33 @@ LABEL org.opencontainers.image.source="https://github.com/heypinchy/pinchy" \
 ENV COREPACK_HOME=/app/.cache/corepack
 RUN corepack enable
 
+# LibreOffice: the knowledge base converts page-shaped Office documents
+# (.doc/.docx/.ppt/.pptx) to PDF before indexing them (#936, lib/knowledge/
+# office-convert.ts). This is the only toolchain that reads the LEGACY binary
+# formats — 13 of the 19 Office files in the reference corpus are .doc/.ppt,
+# and the entire modern parsing ecosystem (Docling, MarkItDown, Unstructured,
+# python-docx/python-pptx) is OOXML-only. Conversion also produces the single
+# renderable artifact the citation chain needs, so a citation can never point
+# at something the reader cannot open.
+#
+# Measured on node:22-slim (247 MB): writer+impress+fonts-liberation = 669 MB
+# (+422 MB); the full `libreoffice` metapackage would be 772 MB. We install
+# only the two page-shaped filters:
+#   - libreoffice-calc is DELIBERATELY absent. A sheet is not a page, and
+#     spreadsheets take a different path entirely (#937/#940).
+#   - fonts-liberation is not optional. Without a metric-compatible font set
+#     LibreOffice substitutes, re-flows, and can drop glyphs — and the corpus
+#     is full of Central/Eastern European diacritics (GODIČ TORKAR, Domžale).
+#   - catdoc (~1 MB) ships catdoc/catppt, the INDEPENDENT text extractors the
+#     conversion's lower-bound word check compares against for the legacy
+#     formats. An oracle derived from LibreOffice would agree with LibreOffice
+#     about text it dropped and verify nothing.
+# `--no-install-recommends` matters here: the recommends pull in a JRE.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+       libreoffice-writer libreoffice-impress fonts-liberation catdoc \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /app
 
 EXPOSE 7777
@@ -213,7 +240,12 @@ HEALTHCHECK --interval=30s --timeout=5s --start-period=10s --retries=3 \
 
 # Create non-root user (entrypoint drops privileges after fixing volume permissions)
 RUN groupadd -r -g 999 pinchy && useradd -r -u 999 -g pinchy -d /app pinchy
-RUN mkdir -p /app/secrets /openclaw-config/workspaces && chown -R pinchy:pinchy /app /app/secrets /openclaw-config/workspaces
+# /var/cache/pinchy-kb-artifacts is the converted-Office-PDF store (#936). It
+# is a named volume in docker-compose.yml precisely so nothing is ever written
+# next to the original: /data is mounted read-only and the docs state that as a
+# product promise. Mirrors the openclaw container's pinchy-pdf-cache.
+RUN mkdir -p /app/secrets /openclaw-config/workspaces /var/cache/pinchy-kb-artifacts \
+    && chown -R pinchy:pinchy /app /app/secrets /openclaw-config/workspaces /var/cache/pinchy-kb-artifacts
 
 COPY entrypoint.sh /entrypoint.sh
 # Plugin-source sync helper the entrypoint runs on every boot. Extracted from
--- a/Dockerfile.pinchy.dev
+++ b/Dockerfile.pinchy.dev
@@ -5,6 +5,16 @@ FROM mirror.gcr.io/library/node:22-slim
 
 RUN corepack enable
 
+# Same LibreOffice toolchain as the production image (see Dockerfile.pinchy for
+# why these exact packages). Dev needs it too: without it, indexing an Office
+# document in the dev stack reports every file as "converter did not run"
+# infrastructure failures, which looks nothing like the production behaviour it
+# is supposed to reproduce.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+       libreoffice-writer libreoffice-impress fonts-liberation catdoc \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /app
 
 COPY pnpm-workspace.yaml package.json pnpm-lock.yaml ./
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -23,6 +23,14 @@ services:
       # land in a separate store), which also keeps the corpus tamper-proof
       # from the web tier.
       - pinchy-data:/data:ro
+      # Converted-Office-PDF artifacts (#936). A separate volume because /data
+      # above is read-only and must stay that way: the converted PDF is what
+      # the index and the preview both point at, and writing it beside the
+      # original would break the corpus's tamper-proof promise. Same shape as
+      # the openclaw container's pinchy-pdf-cache. Safe to delete — every
+      # artifact is re-derivable from its source, keyed by content hash and
+      # converter format version.
+      - pinchy-kb-artifacts:/var/cache/pinchy-kb-artifacts
     restart: unless-stopped
     depends_on:
       db:
@@ -125,6 +133,7 @@ volumes:
   pinchy-secrets:
   openclaw-extensions:
   pinchy-pdf-cache:
+  pinchy-kb-artifacts:
   openclaw-secrets:
     driver: local
     driver_opts:
--- a/docs/src/content/docs/guides/customizing-deployment.mdx
+++ b/docs/src/content/docs/guides/customizing-deployment.mdx
@@ -129,6 +129,35 @@ DB_CPUS=0.5
   raise the limit before assuming something else is wrong.
 </Aside>
 
+### Memory when the knowledge base indexes Office documents
+
+Word and PowerPoint files take a detour on their way into the knowledge base: Pinchy converts them to PDF first, with LibreOffice running inside the `pinchy` container. That's what lets it read the legacy `.doc` and `.ppt` formats at all — no modern parsing library can — and it gives every citation one rendered document that the index and the preview agree on.
+
+Two things about that are worth knowing when you size the container.
+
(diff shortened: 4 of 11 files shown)


clemenshelm:
## Review-Nachtrag (`e0bf01256`)

Ich habe den Code beim Review nicht nur gelesen, sondern gegen echtes LibreOffice gefahren. Zwei Defekte, die der Fake strukturell nicht zeigen kann, plus drei Lücken:

**Der Timeout verwaiste den echten Konverter.** `soffice` exect `oosplash`, das `soffice.bin` *forkt* — der SIGKILL traf nur den Wrapper:

```
kill -9 2991 (oosplash)
--- survivors 3s later ---
3011 ?  Dl  0:02 /usr/lib/libreoffice/program/soffice.bin --convert-to pdf ...
```

Der echte Konverter lief weiter: hielt den Speicher, den der Timeout freimachen sollte, schrieb in ein Staging-Verzeichnis, das der `finally`-Block gerade löschte, und war beim Start des nächsten Batches noch resident. Unter Speicherdruck wird aus einem Timeout eine Kaskade. Jetzt `detached: true` + Kill der Prozessgruppe — in beiden Richtungen im Runtime-Image verifiziert (pid-Kill verwaist, Gruppen-Kill räumt ab).

**Jeder Dokumentfehler zitierte die falsche stderr-Zeile.** Das Image hat absichtlich kein JRE, also druckt LibreOffice bei *jedem* Lauf `Warning: failed to launch javaldx` als erste stderr-Zeile. `summarise()` nahm „die erste nicht-leere Zeile" — also klebte eine permanente, harmlose Warnung an jedem gescheiterten Dokument und verdeckte die einzige aussagekräftige Zeile. stderr ist zudem batch-weit und nennt keine Datei (verbatim: `Error: source file could not be loaded`).

Drei weitere, im Geist derselben Trennung:

- **Env-Leak:** Der Konverter erbte `DATABASE_URL` und `ENCRYPTION_KEY`. Er parst Dokumente, die nicht der Betreiber geschrieben hat, in Formaten mit langer Memory-Safety-Historie → Allowlist. Verifiziert, dass `PATH` + `HOME` konvertiert und die Diakritika intakt bleiben.
- **Voller Store warf aus `convertOfficeFiles` heraus** und verwarf die Ergebnisse bereits fertiger Batches. ENOSPC/EACCES ist genau der Engpass, für den `infrastructure` existiert.
- **Orakel-stdout unbegrenzt:** ein `.doc`, das `catdoc` endlos ausgeben lässt, hätte den Web-Prozess mitgenommen.

**Abdeckung.** Die CI-Logs dieses PRs zeigen `office-convert.libreoffice.test.ts (2 tests | 2 skipped)` — `runSoffice` war also nirgends abgedeckt, wo ein PR es sieht. `office-convert.adapter.test.ts` fährt argv, Private-Profile-Flag, Env-Hygiene, Exit 137, stderr-Deckel und den Timeout-Kill inklusive Enkelprozess gegen Stub-Binaries: braucht einen Prozess, kein 422-MB-LibreOffice, läuft bei jedem PR.

Gesamtzahl damit **51 Tests** in vier Dateien (nicht 38 wie oben) — 49 laufen in CI, 2 bleiben LibreOffice-gated.

### Zwei Punkte für #938 statt hier

- `sweep()` und `pruneOtherFormatVersions()` ruft niemand auf. Die Invalidierung beim Format-Bump ist konstruktiv korrekt, die *Rückgewinnung* nicht: eine Version-Erhöhung lässt die alte Generation dauerhaft auf dem Volume liegen.
- `collectOne` schickt jedes konvertierte PDF durch pdfjs, und #938 wird es für den Text erneut parsen — bei einem großen Korpus die teuerste Operation, zweimal.

Issue #901, opened 2026-07-23
Flaky: kb-attribution Layer-2 self-test — agent not dispatchable within deadline (config.apply stuck in file-watcher debounce)


clemenshelm:
## Symptom

`e2e/integration/kb-attribution.spec.ts:179` — *KB Eval Harness — Layer-2 attribution self-test › well-formed answer passes every grader* — failed in the Integration-Tests shard (2/2) on PR #897 with:

```
Error: OpenClaw runtime did not see agent 615a804a-3611-4539-bff2-62bdbe8f6dd4 as dispatchable within deadline — config.apply likely stuck in file-watcher debounce
```

Failing run: https://github.com/heypinchy/pinchy/actions/runs/29990887558/job/89153859200

## Context

- PR #897 touches only two web component test files — the failure is unrelated to the diff (pre-existing infra/timing flake).
- Distinct from the `pinchy-files` "registers pinchy_ls and pinchy_read" timeout flake noted in the PR #897 description — do not conflate.
- Likely related to the dispatch-probe readiness discussion in #465: the error message itself points at `config.apply` sitting in the file-watcher debounce while the probe deadline runs out.

## Next steps

- [ ] Check how often this specific error fires across recent Integration-Tests runs (same error string).
- [ ] If frequent: either extend the dispatch-probe deadline for this spec or flush/bypass the file-watcher debounce in the test harness (an explicit config.apply readiness signal would fix the class, not the instance — see #465).

clemenshelm:
Not reproducible any more — fixed the day after this was filed, by work that never referenced it.

**The error in this issue is `waitForAgentDispatchable`'s** (`e2e/shared/dispatch-probe.ts:172`). `ensureAgentDispatchable` catches that and always rethrows its own message, which carries `after N recovery attempt(s)`. The two are distinguishable by shape, and the one reported here has no attempt count.

| | |
|---|---|
| Flake fired | 2026-07-23, PR #897 |
| `ensureAgentDispatchable` written | 2026-07-24, `ebff473e0` — "recover KB dispatch setup from OpenClaw's config.apply rate limit" |
| Bounded by an overall wall-clock budget | 2026-07-25, `506e0fde8` |
| `kb-attribution.spec.ts` today | calls `ensureAgentDispatchable`, does not import the plain helper |

So the spec structurally cannot produce this error today.

[...]

Issue #869, opened 2026-07-21
KB Eval Harness Layer 3 never ran: chain of latent bugs (id-inserts ✅, CSRF-export 403 ✅, ANY/ALL SQL, custom-agent methodology)


clemenshelm:
## Summary

The **KB Eval Harness Layer 3** (groundedness sweep, `packages/web/eval/kb/kb-eval-models.spec.ts`) was shipped as "code-complete + unit-tested", but its live dry-run was deferred (disk-blocked), so **it had never actually executed against a running stack**. Its first real run (2026-07-21, validating the #715 native-embeddinggemma retrieval end-to-end) surfaced a chain of latent bugs and one methodology gap.

Layers 1 (retrieval, deterministic HARD gate) and 2 (attribution, deterministic HARD gate) are unaffected and remain green in CI. This issue is scoped to the **stochastic Layer-3 sweep** (`pnpm kb-eval:models`), which is not a CI gate.

## Findings (in the order the first run hit them)

1. **✅ FIXED — id-less corpus INSERTs.** `seedSyntheticCorpus` inserted into `kb_documents` and `kb_chunks` without an `id`. Both columns are `text` with **no DB default** — the app supplies them via Drizzle `$defaultFn(() => crypto.randomUUID())` — so the raw seed hit `null value in column "id" ... violates not-null constraint` on the very first insert. Fixed by minting the id in SQL with `gen_random_uuid()::text` (Postgres 13+ builtin).

2. **✅ FIXED — diagnostics-export 403 (CSRF).** `getRawAssistantMessage` (the raw-answer capture, whose own doc comment flagged "NEEDS VALIDATION AGAINST THE RUNNING STACK, read from source, not observed live") POSTs to `/api/diagnostics/export` via `page.request.post` with **no `Origin`/`Referer` header**. The CSRF gate (`src/server/csrf-check.ts`) rejects that as `missing-origin-and-referer` → 403, so every answer capture failed and every run recorded `run-infra-error`. Every other mutating eval call sets `Origin: PINCHY_URL` via `run-eval.ts`'s `mutatingHeaders`; this one capture bypassed that helper. Fixed by sending an `Origin` matching the page's host.

   > Both fixes: see the linked PR. With them applied the pipeline runs end-to-end and **grades for the first time ever**.

[...]

Issue #947, opened 2026-07-28
dev-workflow: cut the per-change latency agents pay (pre-push build, jsdom-by-default, gate placement)


clemenshelm:
**Handover issue.** Written from observations during an unrelated session on 2026-07-28; not investigated to a conclusion. Goal: **a faster agent workflow — fewer wasted minutes, fewer wasted tokens, without giving up any real gate.**

## Why this is worth a session

Agents pay these costs on *every* iteration, and unlike a human they cannot "go get coffee" — a blocked command is a burned tool call, sometimes a burned timeout and a retry. The costs have crept in one at a time and nobody has looked at the total.

The trigger: `git push` **exceeded a 5-minute timeout**, had to be moved to a background job, and the developer was unaware a build even ran on push.

## Measured (2026-07-28, real observations)

| step | cost | how often |
|---|---|---|
| `pre-push` → `pnpm install --frozen-lockfile` + `pnpm build` (`next build --turbopack`) | **> 5 min** under load; minutes when idle | **every push** |
| `pnpm test` (web) | 157–160 s idle, 306–412 s under load | every verification |
| `pnpm test:db` | 212 s | every verification |
| `pnpm install --frozen-lockfile` (fresh worktree) | 11–21 s | per worktree |
| `pre-commit` → drizzle snapshot check + `lint-staged` (eslint on staged web files, prettier on all staged) | seconds | every commit |

## Lead 1 — jsdom for everyone, needed by 22 %

`packages/web/vitest.config.ts` sets `environment: "jsdom"` globally.

```
Test files total:      654
with any DOM usage:    143
paying for nothing:    511   (78 %)
```

And in a measured run, environment setup was **the single largest bucket** — bigger than running the tests:

```
Duration 157.38s (transform 42.78s, setup 128.07s, import 284.99s,
                  tests 304.44s, environment 1155.76s)
```

(Buckets are summed across workers, wall clock is 157 s — but the ratio is the point.)

Direction to evaluate: default `environment: "node"`, opt into jsdom per file via `// @vitest-environment jsdom` or a path-based `environmentMatchGlobs`. **Verify before believing** — the 143-file count comes from a crude grep for `@testing-library` / `document.` / `window.`, which will both over- and under-count.

## Lead 2 — a full `next build` on every push

The `pre-push` hook runs `pnpm build`. It is not worthless: `next build` catches client/server boundary errors that `tsc --noEmit` does not, and that class has bitten this repo before. But:

[...]

Issue #543, opened 2026-06-18
RFC/Tracking: Pinchy Skill Layer (consolidates #354/#126/#372)


clemenshelm:
## Vision

Pinchy adopts the OpenClaw-native skill layer as the foundation for **composable agent capabilities** — instead of growing the template list to cover every plugin combination, agents are assembled from three layers:

```
Layer 1: Template       Persona + default skill set
                        ("Smithers", "Market Monitor", "Lead Researcher")
Layer 2: Skill          SKILL.md with workflow + required capabilities
                        ("web-search", "crm-read", "outreach-drafting")
Layer 3: Capability     Concrete tool providers
                        (pinchy-odoo, pinchy-files, pinchy-web, MCP servers)
```

This consolidates three existing RFCs (#354, #126, #372) into one coherent direction and unlocks **hybrid templates** (Lead Researcher = Web + Odoo, etc.) which are currently impossible because `pluginId: string | null` is single-plugin.

## Why now

- **OpenClaw 2026.6.5** (bundled in Pinchy v0.6.0) implements the [AgentSkills.io](https://agentskills.io) spec natively: `SKILL.md` with YAML frontmatter, six-tier loading precedence (`<workspace>/skills` > `.agents/skills` > `~/.agents/skills` > `~/.openclaw/skills` > bundled > extra-dirs), per-agent allowlist via `agents.list[].skills`, gating via `metadata.openclaw.requires`, plus Skill Workshop and ClawHub. We get the composer/loader/allowlist mechanics for free.
- **Industry alignment** — Anthropic Agent Skills, OpenAI Codex Skills, Glean Skills, GitHub Copilot all converge on the same SKILL.md format. Building on this means Pinchy ecosystem-imports work later without a parallel system.
- **MCP scaling pain is documented** — at ~30 tools per agent, tool-selection accuracy degrades; at 100+ it breaks. ClawHub already has 13k+ community skills. Without a skill layer, the MCP marketplace would flood agents with tool definitions and degrade quality.
- **Template-explosion is the alternative** — `pinchy-web` is the last external plugin without templates today. Adding 4 hybrid templates per integration (CRM, ERP, ...) takes us toward an unmaintainable matrix.

## Architecture (validated by smoke-test against OC 2026.6.5)

[...]

PR #1003, opened 2026-07-30
fix(eval): make the Layer-3 KB sweep produce a number worth reading

clemenshelm:
Closes both open items in #869. Together they are what stands between the sweep and its first trustworthy measurement — one made it fail, the other made it lie.

Both landed on one branch on purpose: split, the second would have to stack on the first, and they are two halves of the same claim ("this sweep measures groundedness").

---

## Item 3 — the premise lookup never returned a row

`fetchChunkTexts` built its filter as `= ANY(${sql.array(sourcePaths)})`. postgres.js already infers `text[]` from a string array; `sql.array` produces a parameter Postgres rejects there:

```
PostgresError: op ANY/ALL (array) requires array on right side
```

Reproduced against a real Postgres before changing anything — both spellings, two elements and one:

```
❌ sql.array(paths)  [current]        op ANY/ALL (array) requires array on right side
✅ direct interpolation               2 rows
❌ sql.array, single element          op ANY/ALL (array) requires array on right side
✅ direct, single element             1 row
```

**It is not intermittent — the observation was.** #869 records it firing "on some runs", which points triage at a data-dependent trigger. There isn't one. The function returns early on an empty path list, and most answers in that first sweep cited nothing resolvable, so most runs never reached the SQL. The runs that did were the ones whose Sources list parsed cleanly: the answers the sweep exists to grade. They were filed as `run-infra-error` and dropped from the scorecard.

Worth carrying into the next triage: **an early return in front of a broken statement turns "always" into "sometimes."**

The query moves to `eval/kb/chunk-texts.ts` so a DB-backed test can reach it — deliberately not into `kb-eval-shared.ts`, which imports `@playwright/test`; dragging Playwright into the vitest integration lane to test one SQL statement is a bigger dependency than the thing under test. It takes `orgId` as a parameter instead of closing over `DEFAULT_ORG_ID`, which is what lets the test assert org scoping with a tenant of its own.

`kb-eval-chunk-texts.integration.test.ts` covers multi-path lookup, the single-path shape a minimal Sources list produces, org scoping (another tenant's identical path must not ground an answer), and the empty-input early return. It has to be an integration test: this is a serialization fault between driver and database, and **no mock can see it**. Nothing about the TypeScript was wrong, which is how it survived review and a full unit suite.

## Item 4 — the sweep graded a contract nobody had stated

The sweep created its agent with `templateId: "custom"`. That template's `defaultAgentsMd` is **`null`** — no cite-then-answer rule, no Sources list, no closed-set constraint — and then the run was graded against precisely that contract. kimi-k2.6 and qwen3.5:397b scored `passRate 0`, dominated by `ungrounded-claim` and `citation-unresolved`.

That measures whether a model invents an unstated contract. No Pinchy user depends on it, because every KB agent ships with the instructions. The sweep now creates a **`knowledge-base`** agent, so it measures what we ship.

**The guard is the interesting half.** Asserting `KB_SWEEP_TEMPLATE_ID === "knowledge-base"` would pin a string and miss the point: the contract could be edited out of the template tomorrow and the sweep would keep publishing groundedness numbers for an agent that was never told to cite. So the test renders the template's AGENTS.md and checks the clauses the graders enforce — one assertion per grader premise:

| Assertion | Grader premise |
|---|---|
| inline `[N]` citations | `citation-unresolved` |
| a Sources list, rendered as bullets | `source-uncited`, `sources-format` |
| inline ↔ list match "no more and no fewer" | both directions |
| document path + position per entry | `path-not-cited` |
| answer only from the returned sources | `ungrounded-claim` |
| the corpus root is rendered | the agent can find its documents |

Verified red before green: with the constant still on `"custom"`, **8 of 9 fail**. The 9th is the discrimination test, which asserts `custom` renders no instructions at all — the property that made the original setup silently wrong.

---

## Gates

- `format:check`, `lint` (0 errors), `typecheck` — green
- Full unit suite: **738 files, 10 422 tests** green
- `src/__tests__/lib/knowledge/` integration folder against a real pgvector Postgres: **6 files, 70 tests** green, including the Layer-1 retrieval gate

## Also corrected

The spec's file header still listed `fetchChunkTexts` under "NEEDS VALIDATION AGAINST THE RUNNING STACK". It is validated now — and it had been broken for the entire time that note stood, which is the argument for validating rather than noting.

## Noted, not changed

`setupKbSweepAgent` still uses the non-recovering `waitForAgentDispatchable`, while `kb-eval-shared.ts` next door uses `ensureAgentDispatchable`. Same class as #901. Out of scope here; worth a look when the sweep starts running regularly.

## What this does not claim

The sweep still has not produced a scorecard. This makes it *able* to — the premise material loads and the agent is instructed to do what is graded. The first real run comes next, and per the operational note in #869, `eval/kb/results/` must be cleared first so pre-fix `run-infra-error` rows do not mix into it.

### DIFF
--- a/packages/web/eval/kb/chunk-texts.ts
+++ b/packages/web/eval/kb/chunk-texts.ts
@@ -1,0 +1,48 @@
+/**
+ * The groundedness premise lookup for the Layer-3 sweep, extracted from
+ * `kb-eval-models.spec.ts` so a DB-backed test can reach it.
+ *
+ * It lives outside `kb-eval-shared.ts` on purpose: that module imports
+ * `@playwright/test`, and pulling Playwright into the vitest integration lane
+ * to test one SQL statement would be a much larger dependency than the thing
+ * under test.
+ */
+
+/**
+ * Fetches every chunk's `chunk_text` for the given `sourcePaths`, keyed by
+ * path — the groundedness premise material for whichever sources the answer's
+ * Sources list actually cited (`citedSourcePaths`, not the full retrieved set
+ * — see that function's doc comment for why). A direct SELECT against the
+ * already-seeded corpus, not a re-run of `retrieve()`: we already know exactly
+ * which paths were retrieved (from the audit row), so this needs no new
+ * search/embedding call, only the stored text.
+ */
+export async function fetchChunkTexts(
+  dbUrl: string,
+  orgId: string,
+  sourcePaths: string[]
+): Promise<Map<string, string[]>> {
+  const map = new Map<string, string[]>();
+  if (sourcePaths.length === 0) return map;
+
+  const { default: postgres } = await import("postgres");
+  const sql = postgres(dbUrl);
+  try {
+    // Interpolate the array directly. `sql.array(paths)` reads like the more
+    // explicit spelling and is the one that does not work: postgres.js already
+    // infers `text[]` from a string array, while `sql.array` produces a
+    // parameter Postgres rejects on the right side of ANY (#869).
+    const rows = await sql<{ source_path: string; chunk_text: string }[]>`
+      SELECT source_path, chunk_text FROM kb_chunks
+      WHERE org_id = ${orgId} AND source_path = ANY(${sourcePaths})
+    `;
+    for (const row of rows) {
+      const texts = map.get(row.source_path) ?? [];
+      texts.push(row.chunk_text);
+      map.set(row.source_path, texts);
+    }
+    return map;
+  } finally {
+    await sql.end();
+  }
+}
--- a/packages/web/eval/kb/kb-eval-models.spec.ts
+++ b/packages/web/eval/kb/kb-eval-models.spec.ts
@@ -209,9 +209,11 @@ function seedNoackCorpus(): never {
  * Creates a fresh **Knowledge Base** agent scoped to `knowledge_search` + the
  * eval corpus root, waits for it to be dispatchable.
  *
- * The template is the measurement, not a detail: it is what carries the
- * cite-then-answer instructions the graders enforce. See `sweep-agent.ts` for
- * why a bare `custom` agent made the first sweep unreadable (#869 item 4).
+ * The template is the measurement, not a detail: it is what brings in the
+ * cite-then-answer instructions the graders enforce — today through the
+ * `knowledge-search` skill it declares, which agent creation copies onto the
+ * agent row. See `sweep-agent.ts` for why a bare `custom` agent made the first
+ * sweep unreadable (#869 item 4).
  */
 async function setupKbSweepAgent(cookie: string): Promise<{ agentId: string }> {
   const createRes = await pinchyPost(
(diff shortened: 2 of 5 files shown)


clemenshelm:
## Review

Two real defects, both correctly diagnosed, both pinned by a test that fails without the fix. I reproduced the `sql.array` fault myself against a real Postgres (`op ANY/ALL (array) requires array on right side`, 3 of 4 tests red) — the integration test earns its keep, and the reasoning about why it *had* to be DB-backed is right: nothing about the TypeScript was wrong.

The framing in item 3 is the part worth keeping: **an early return in front of a broken statement turns "always" into "sometimes."** That is a triage lesson, not a bug note.

### One blocking issue — fixed on the branch

The guard read only `generateAgentsMd(template)`. That was true when it was written and stale by the time it landed: #543/#544 moved the retrieval and citation workflow out of `defaultAgentsMd` into the `knowledge-search` SKILL.md, leaving the KB template with persona only. Rebased on main, **5 of 9 assertions went red** — `document path`, `position`, `no more and no fewer`, `only using the numbered sources`, `never fabricate` — against a contract that had not been weakened at all. That is what the failing `quality` job was.

The interesting part is *why* that is a guard defect and not just a stale test. A guard that cannot tell "the contract moved" from "the contract is gone" gets loosened the first time it blocks a legitimate refactor, and the day it is loosened is the day it stops catching the thing it exists for. So it now reads the union of AGENTS.md and every body in `defaultSkills`, and stays red only when a clause is in neither source.

Verified in both directions:
- as-is on main: 9/9 green;
- `defaultSkills: []` on the KB template: 5 red again.

Two clause assertions now follow the skill's actual wording (`only from the returned passages`, `never cite a number that wasn't in the returned list`) — the same premise `ungrounded-claim` grades. The discrimination test asserts `custom` renders nothing from **either** source.

I checked the delivery chain rather than assuming it: `/api/agents` copies `defaultSkills` onto the agent row (`agents-create.test.ts` pins `skills: ["knowledge-search"]` for this exact template), the spec's follow-up PATCH touches only `allowedTools` and `pluginConfig`, and `build.ts` materializes from `agent.skills`. The sweep agent does get the contract. The guard deliberately does not re-assert that hand-off — it is pinned one level down, and a second copy would decay out of step.

### One smaller fix

`const DB_URL = process.env.DATABASE_URL ?? ""` — postgres.js reads an empty connection string as "use libpq defaults", so a missing env var would have pointed the test at whatever Postgres the host happens to run, including another worktree's, instead of saying what went wrong. Same failure shape the `db-url.ts` comment describes. It throws now.

### Noted, not changed

- **Item 5 of #869 stays open** (`gpt-oss:120b` model-pin 400). The body says "closes both open items"; that is accurate for the two blocking ones, but the issue should not be closed on this PR — item 5 and the `results/` clearing note are still live.
- `setupKbSweepAgent` still uses the non-recovering `waitForAgentDispatchable` — agreed, out of scope, same class as #901.
- Both doc comments that said the *template* carries the contract now say the skill does. Left stale, they would send the next reader to the file the clauses just left.

### Gates (local, Node 22)

- `format:check`, `lint` (0 errors), `typecheck` — green
- `pnpm test:scripts` — 675 green
- Full unit suite — **743 files, 10 505 tests** green
- `pnpm test:db` against a real pgvector Postgres — **53 files, 399 tests** green

Two aborted runs before that one were environment, not code, and are worth recording because both cost real time: `better-sqlite3` was compiled for Node 24 in this worktree (24 failures across `pinchy-files` reading as ordinary assertion errors, the failure mode #947 describes), and `openai-compatible-discovery.test.ts` timed out once under load 37 and passes in isolation.

PR #1010, opened 2026-07-31
fix(audit): type integration.credentials_tested as the route writes it

clemenshelm:
## What does this PR do?

The `AuditLogEntry` union member for `integration.credentials_tested` did not describe the rows that are actually written.

- Declared: `{ imapHost, smtpHost, reason?, emailHash?, emailPreview? }`
- Written by `POST /api/integrations/imap/test`: `{ imapHost, smtpHost, imapCode?, smtpCode?, ...redactEmail(username) }`

So `reason` was advertised and never written, and the per-leg failure codes — the entire diagnostic value of a failure row — were written and never declared. The type was documentation that was quietly wrong: someone learning the shape from it would not discover that a failed probe records *which* leg failed and *how*.

It typechecked because the route built its entry as an inferred local and passed the variable to `appendAuditLog`, which skips TypeScript's excess-property check.

Related to #1005, which documented these fields in `docs/src/content/docs/concepts/audit-trail.mdx`. This is the code side of the same mismatch. No docs change is needed here — the docs already describe the runtime shape.

### Closing the hole took more than the obvious annotation

Both steps were verified with a canary (add a bogus field, confirm `tsc` rejects it) rather than by reading the rules:

- **Annotating `auditEntry: AuditLogEntry` does not help.** Excess-property checking applies to object *literals* being assigned, and `detail: auditDetail` hands over a variable. The canary passed cleanly. The annotation therefore sits on `auditDetail`, with its type pulled out of the union via `Extract` so the two can't be maintained apart.
- **A conditional spread is invisible to the check too.** `...(imap.ok ? {} : { imapCode: imap.code })` would have left exactly the hole the annotation exists to close — for the two fields this PR is about. Both codes are now direct properties (`imap.ok ? undefined : imap.code`); `undefined` drops out on JSON serialization, so the stored row is unchanged.

`imapCode`/`smtpCode` are typed as `ProbeFailureCode` rather than `string`, so the reader also learns which codes exist. The import is `import type`, and the real `pnpm build` confirms the IMAP probe's runtime deps stay out of `audit.ts`'s module graph.

### Tests

- The type-level assertion in `audit-types.test.ts` covers the *other* direction: a field the type declares that nothing writes (which is what `reason` was). It was red before the fix.
- The route tests move from field-by-field to whole-shape `toEqual` — a per-field assertion cannot see a stray key — across all four combinations: success (no code), IMAP failure, SMTP failure, both legs blocked.
- New case: a non-email-shaped username, where the optional redaction pair is absent entirely.
- Mutation check: forcing `imapCode` unconditionally turns 4 tests red.

## Type of change

- [x] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [x] 🧪 Tests
- [ ] 🔧 Tooling / CI

## Checklist

- [x] I've updated the documentation (if applicable) — none needed, see above
- [x] All existing tests pass

## Verification

Run locally on Node 22, all green: `format:check`, `lint` (0 errors), `typecheck` (incl. test files), `pnpm test` (10507 passed), `test:db` (399 passed), `test:scripts`, `typecheck:plugins`, `test:plugins`, `pnpm build`.

One thing worth flagging rather than hiding: `pnpm test:scripts` reported `fail 1` on its very first run and was green on four subsequent runs under identical parallel load. I did not capture which test it was and could not reproduce it; nothing here touches `scripts/`. Plausibly one of the port-allocation guards probing real ports while other worktrees' `*-testdb` containers were bound.


clemenshelm:
Rebased onto `main` (the branch predated the new `check-docs-required` guard) and added a docs commit.

**Why the guard fired, and what I did about it:** it flags any change to an API route or the audit event catalogue that touches no `docs/` file. The audit-trail row for `integration.credentials_tested` was already correct — #1005 landed it, and it already names `imapCode`/`smtpCode` and no `reason`. So that half needed nothing.

Reading `reference/api.mdx` against the route (via the `review-docs` skill) did turn up a real gap, and it is the same gap this PR is about. `POST /api/integrations/imap/test` documented its request body and nothing else, in a reference where 68 other endpoints document their response. Two things a reader would get wrong:

- **The route answers `200` whether the probe passed or failed.** A rejected login is a result, not a request error — an API consumer has to read `ok`, not the status code. That was written down nowhere.
- **`imapCode`/`smtpCode` appear in the audit-trail reference with no enumeration of their values anywhere in the docs.** Someone filtering audit rows on `imapCode` could not find out what to filter for. The codes are now listed where they originate, with the audit trail named as the second place they surface — which is exactly the connection this PR makes in the type.

Also documents the two suggestion outcomes beyond a port switch (`all_smtp_blocked`, and a `null` suggestion when the port you're already on is the only reachable one) — both verified against `buildSuggestion`, along with the `blocked` code, which is broader than "private address" and now links to the existing guide section instead of restating it.

Re-verified after the rebase on Node 22: `pnpm test` (10507 passed), `typecheck`, `test:scripts` (754 passed), `format:check`, and the docs gates `build` + `check:anchors` + `check:tables`. `check-docs-required` is green.

clemenshelm:
**CI is red on a test this PR does not touch.** Filing rather than re-running, since a re-run would have hidden it.

`Integration Tests (1/2)` failed on `19-durable-agent-error-banner.spec.ts:72` — the duplicate-write confirm never opened after clicking Retry. `main` is green, so it is intermittent.

The obvious explanation is wrong, which is why it was worth the artifact download: the integration suite shares one Smithers session, so the natural guess is that the click landed on a stale error bubble from an earlier test. The page snapshot shows exactly one `Retry` button, in this test's own bubble. The click was correct; the dialog genuinely did not appear.

The cause is a race in `client-router.ts`. The retry gate comes from `agentRanToolSince()`, which asks whether a `tool.%` audit row exists for the run — OpenClaw doesn't signal tool execution as a chunk, so the audit trail is the only signal. But that row is written by the plugin over a separate HTTP call, asynchronous to the chunk stream. When the error chunk wins the race, `sideEffects` is `false` and Retry is ungated.

That is a real product gap, not just a flaky assertion: a run that already applied writes can offer an unguarded Retry. Filed as #1013 with the analysis and three possible directions. Out of scope here — this PR is a type correction and its own tests are green.

The three review fixes from `/review` are in `d2457d3c1`; all other checks on that commit pass.

Issue #1013, opened 2026-07-31
Race: the duplicate-write retry gate can be missing on the live error frame


clemenshelm:
## What happened

`e2e/integration/19-durable-agent-error-banner.spec.ts:72` — *"a failure after a tool call warns about duplicates and gates retry behind a confirm"* — failed on an unrelated PR ([#1010](https://github.com/heypinchy/pinchy/pull/1010), a type-only audit fix that touches none of this code). `main` is green, so it is intermittent.

```
Error: expect(locator).toBeVisible() failed
Locator: getByRole('alertdialog')
  > 97 |     await expect(liveConfirm).toBeVisible();
```

Run: https://github.com/heypinchy/pinchy/actions/runs/30607853643/job/91085401465

## What it is not

The obvious explanation — the integration suite shares one Smithers session, so the click landed on a stale error bubble from an earlier test — is **wrong**. The failure artifact's page snapshot shows exactly one `Retry` button, in the bubble for this test's own message (`E2E_TOOL_THEN_RATE_LIMIT: do the thing`). The earlier test's error renders as history text with no Retry. The click hit the right button; the dialog simply never appeared.

## The actual cause

The retry gate is driven by `liveSideEffects`, computed in `client-router.ts` when the error chunk arrives:

```ts
const sideEffects = await agentRanToolSince(args.agent.id, args.runStartedAt);
```

`agentRanToolSince` (`server/chat-session-errors.ts:39`) asks whether a `tool.%` audit row exists for this agent since the run started. OpenClaw does not signal tool execution as a chunk, so the audit trail is the only signal — but that row is written by the **plugin**, through a separate HTTP call back into Pinchy, entirely asynchronous to the chunk stream. If the error chunk arrives before that row commits, `sideEffects` is `false`.

So the ordering is: tool executes → error chunk races the tool's audit write → whichever wins decides whether the user is warned.

## Why this matters beyond the test

The failure mode in production is that a run **which already applied writes** offers an ungated Retry. The user re-sends and the writes are applied twice — precisely what the confirm exists to prevent. The test is not flaky about something cosmetic; it caught the gate being absent.

Note the deliberate fallback in `persistDurableChatError`'s catch block does the same thing (`return false`, "so the in-chat retry stays ungated rather than spuriously demanding a duplicate-write confirm"). That is a stated trade-off for a persist failure. The race is the same outcome arriving unnoticed, which is a different thing.

## Directions

[...]

PR #1014, opened 2026-07-31
test(#1013): rule out one cause of the missing retry gate, and make the next failure decisive

clemenshelm:
## What does this PR do?

Follow-up to #1013 — the intermittent failure of `19-durable-agent-error-banner.spec.ts:72`, where the duplicate-write confirm never opens and Retry re-sends unguarded.

**This does not fix the gate.** The cause is not established, and a fix would be a guess. What it does is remove one candidate permanently and make the next occurrence say which of the remaining ones it was.

The problem worth attacking first is that the run's artifacts cannot answer the question at all: OpenClaw does not log tool executions at the E2E log level, postgres has no statement logging, pinchy has no request logging. Three different bugs produce an identical red.

### Rules out the fake LLM

The spec rests entirely on `TOOL_THEN_RATE_LIMIT` dispatching `pinchy_ls` in round 1 and returning the 429 only in round 2. If it ever went straight to the 429, the run would fail with no side effect — indistinguishable from the bug the spec exists to catch.

Round selection is a pure function of the request's messages, so it needs none of the missing logging. `fake-ollama-tool-then-rate-limit.test.ts` pins it in-process against the real dispatcher, using the histories the integration suite actually produces — its specs share one Smithers session, so by the time this trigger runs the conversation already carries earlier triggers, including `E2E_RATE_LIMIT_ERROR`, whose handler sits directly above this one and would swallow the request if selection keyed off the wrong message.

It doesn't. All four cases pass, including a stale tool result from an earlier turn. That candidate is off the table and stays off it.

### Makes the next occurrence decisive

`describeToolAudit()` asks the audit trail what it knows, while the stack is still up, and carries the answer into the assertion message. The three answers are three different bugs:

| Answer | Meaning |
| --- | --- |
| no `tool.pinchy_ls` row at all | the tool never ran — `sideEffects: false` was **correct** and the fault is upstream of the gate |
| row exists under a different `resource` | audited but not attributable to this agent, so `agentRanToolSince()` can never find it |
| row present and matching | it existed and the lookup missed it — a race with the plugin's audit POST |

The helper never throws: a diagnostic that fails would take the real failure's message with it.

### Recorded so nobody redoes it

- The `agent:unknown-agent` fallback in the tool-use route is reachable, but only when **both** `agentId` and `sessionKey` are absent, which a chat run does not do. Investigated, not established as the cause.
- I earlier claimed OpenClaw's log proved the tool never ran. **That was wrong** — the tool names in that log come from a config warning about *unknown* allowlist entries, so a known tool is absent by definition. The claim is retracted in #1013.

## Type of change

- [ ] 🐛 Bug fix
- [ ] ✨ New feature
- [ ] 📝 Documentation
- [ ] ♻️ Refactor
- [x] 🧪 Tests
- [ ] 🔧 Tooling / CI

## Checklist

- [x] I've updated the documentation (if applicable) — no reader-facing surface moves
- [x] All existing tests pass

## Verification

Node 22: `pnpm test` three consecutive runs green (10509 passed), `test:scripts` (754), `typecheck`, `lint` (0 errors), `format:check`. The new file passes 5/5 in isolation. Playwright compiles and lists the modified spec.

One thing to flag rather than bury: the **first** of those full-suite runs reported a single failure, and the next three were clean. I didn't capture which test it was and cannot name it, so I can't rule out that it was mine — though the new file passed 5/5 isolated and the whole suite passed three times after. Worth knowing if it recurs.

### DIFF
--- a/packages/web/e2e/19-message-hover-layout-shift.spec.ts
+++ b/packages/web/e2e/19-message-hover-layout-shift.spec.ts
@@ -178,8 +178,39 @@ test.describe("message action-bar layout shift", () => {
       return below.y - target.y;
     };
 
+    // Take the baseline only once the gap has stopped moving on its own.
+    //
+    // It moves for ~100ms after mount with no hover involved — measured by
+    // sampling `gap()` eight times at 100ms with the hover removed entirely:
+    // [104.556, 105, 105, 105, 105, 105, 105, 105]. Sampling inside that window
+    // gives a baseline that differs from every later reading, and since the
+    // hover path then waits for the bar to mount, the test reports the
+    // settle as a hover-induced shift. That is what failed in CI (#1016):
+    // "gap 103.894287109375px → 105px" — a slower host, so the baseline landed
+    // further from the settled value and the 1px tolerance broke.
+    //
+    // Waiting is the fix rather than a wider tolerance: the tolerance is what
+    // gives this test its resolution, and the delta it was rejecting was a
+    // measurement artifact, not a layout shift. With the wait, the delta is
+    // exactly 0 (measured across runs), so `< 1` has room to spare.
+    const settledGap = async () => {
+      let previous = await gap();
+      await expect
+        .poll(
+          async () => {
+            const next = await gap();
+            const unchanged = Math.abs(next - previous) < 0.01;
+            previous = next;
+            return unchanged;
+          },
+          { timeout: 5000, message: "the thread's layout never stopped moving" }
+        )
+        .toBe(true);
+      return previous;
+    };
+
     const heightBefore = (await hoverTarget.boundingBox())!.height;
-    const gapBefore = await gap();
+    const gapBefore = await settledGap();
 
     await hoverTarget.hover();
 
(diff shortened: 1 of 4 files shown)


clemenshelm:
Added the fix for the layout-shift flake (#1016) that was failing this PR's `E2E Tests` job — `35341fafe`.

**It carries here rather than in its own PR** because this PR cannot go green without it, and a separate PR would have to be merged first to unblock it. Both commits are the same kind of work: an E2E test that reports something other than what it measured. Stated plainly so it isn't mistaken for scope creep.

**The diagnosis is measured, not assumed.** Sampling `gap()` eight times at 100ms with the hover **removed entirely**:

```
[104.556, 105, 105, 105, 105, 105, 105, 105]
```

The gap settles on its own within ~100ms of mount, with nothing hovering it. The test read its baseline inside that window, then took the hover path — which waits for the action bar to mount, i.e. waits out the settle — and reported the difference as a hover-induced shift. CI is slower, so its baseline landed further out (103.894) and the 1px tolerance broke. With a settle-wait in front, the delta is exactly 0, locally 5/5.

**Widening the tolerance would have been wrong twice.** It was the tempting fix, with precedent in `6937354ec` (which relaxed the height assertion in this same file). But it would have cemented a measurement artifact as expected behaviour, and spent the tolerance that gives the test its resolution. The regression it guards — the action bar landing in flow — moves the gap by the bar's ~34px, so resolution was never the scarce resource; a correct baseline was.

The wait polls for two consecutive identical readings instead of sleeping, per AGENTS.md — a real signal, and it adapts to a host slower than the one I measured on.

Issue #987, opened 2026-07-30
security: credentials endpoint hands out any connection's secrets to any agent


clemenshelm:
## Problem

`GET /api/internal/integrations/[connectionId]/credentials` authorizes **only** on possession of the shared gateway token. It never checks that the calling agent is actually granted that connection via `agent_connection_permissions`.

Because `gatewayToken` is a single shared secret inlined into **every** plugin config (`lib/openclaw-config/build.ts` — same `gatewayTokenString` handed to `pinchy-odoo`, `pinchy-email`, `pinchy-files`, `pinchy-context`, `pinchy-knowledge`, `pinchy-audit`, `pinchy-transcript`), any code running inside the OpenClaw container can request decrypted credentials for **any** `connectionId` in the instance — regardless of which agent that connection is configured for.

## Where

- Handler: `packages/web/src/app/api/internal/integrations/[connectionId]/credentials/route.ts:12-27` — validates the token, reads `connectionId` from the URL, returns decrypted credentials. No agent/connection cross-check.
- Auth helper: `packages/web/src/lib/gateway-auth.ts:11-20` (constant-time compare of the shared token — working as designed, just not sufficient alone).
- Grant table that *should* be consulted: `packages/web/src/db/schema.ts:489-514` (`agent_connection_permissions`: `agentId`, `connectionId`, `model`, `operation`).
- Caller: `packages/plugins/pinchy-odoo/index.ts:190-197` (`fetchCredentials`).

## Why it matters

The three-layer default-deny model (factory returns `null` without agent config → absolute `tools.allow` allowlist → per-operation `checkPermission`) is solid for the **LLM tool-calling** path. This endpoint sits *beside* that model: it's reachable by any code in the OpenClaw process, and it hands out decrypted third-party credentials (Odoo, email) for connections the agent was never granted.

Not a break in the agent sandbox, but a missing defense-in-depth check on a high-value endpoint — and cheap to close.

## Proposed fix

Require the caller to identify the agent (the plugins already know `ctx.agentId`), and verify the `(agentId, connectionId)` pair exists in `agent_connection_permissions` before decrypting. Reject with 403 otherwise, and write a `tool.denied`-style audit row so the attempt is visible.

Follow-up worth considering separately: replace the single shared gateway token with per-plugin or per-agent scoped tokens, so possession alone stops being instance-wide authority.

## Acceptance

- [ ] Credentials endpoint requires an agent id and validates the grant against `agent_connection_permissions`
[...]

Issue #956, opened 2026-07-28
Chat can wait forever for the initial history frame (no deadline, no retry)


clemenshelm:
## Symptom

Reported on the **production** PWA, 2026-07-28. On mobile, several agents were opened one after another; each stayed on the loading indicator indefinitely — no error, no timeout, no retry offered. A reload of the PWA **on the desktop** was followed immediately by the agents appearing on mobile.

## The defect that holds regardless of root cause

**The client's wait for initial history has no deadline.**

`hasInitialContent` (`use-ws-runtime.ts:498`) is what takes the chat out of "starting", and it flips only when a `history` frame arrives — either with messages or with the authoritative `sessionKnown: true` empty signal added for #197. Nothing in `use-ws-runtime.ts` arms a timer for that frame; the timers present (`delayTimerRef`, ack timers, reconnect backoff) all cover *sending*, not the initial load.

So if the frame is lost for **any** reason, the UI parks on the loading indicator forever: no error state, no automatic retry, no user-visible way out except a manual reload. That is a real gap independent of which cause dropped the frame, and it is what turned a transient fault into an indefinite dead end here.

## Why "the server just didn't answer" is unlikely

`ClientRouter`'s history path (`client-router.ts:956-1058`) sends a `history` frame on **every** branch:

- messages found → `history` with messages
- empty but session known → `history` with `sessionKnown: true`
- `waitForConnection()` rejects (10 s, `CONNECTION_TIMEOUT_MS`) → catch → retry once → `history` either way
- OpenClaw not connected and session unknown → `history` with `messages: []`
- otherwise → greeting

There is no path that returns without answering. So the likelier reading is that the frame never reached the device, or the request never reached the server — not that the server declined to send one.

## Candidate causes — none confirmed

1. **Dead-but-unreported WebSocket after PWA resume.** A backgrounded PWA socket can be torn down by the OS without a `close` event reaching the page. The client believes it is connected, sends the history request into a void, and — with no deadline — waits forever. The extensive lifecycle handling (`visibilitychange`, `pageshow`, `freeze`, `resume`, `use-ws-runtime.ts:1518-1522`) exists precisely for this, so if this is it, the gap is in detection, not in intent.

[...]

Issue #934, opened 2026-07-28
kb: let a reader download the cited document, not just view it


clemenshelm:
Part of the Office/scanned-document work. **Wave 1 — parallel with the other Wave-1 issues.**

## Why

Direct customer feedback at the 2026-07-28 demo. On seeing the cited sources:

> "Vielleicht mal anklicken, dass sie gleich noch einen Link finden zu dem Dokument, weil manchmal ist es dann trotzdem so, okay mach die Antwort, hätte aber trotzdem gern das Dokument, dass man den Kunden schicken kann."

Today they reach the file through Citrix: save to a local disk, then attach to mail, because Citrix cannot attach directly. Removing that detour is one of the clearest wins available:

> "Wenn du sie gleich da runterholen kannst, ist es natürlich super."

The viewer opens the document but offers no way to keep it. `PdfDialog`'s header currently has exactly two controls: *Open in new tab* and *Close*.

## What already works

`GET /api/agents/[agentId]/workspace-file` serves with a disposition derived from the extension (`agent-file-content-type.ts`): `application/pdf` → `inline`, **everything else → `attachment`**, which is the anti-XSS control (HTML/SVG under an allowed path must never render inline as same-origin content). So the serving half is done; only the affordance is missing.

## Scope

- A download control in the `PdfDialog` header.
- Build it so a **second** entry is not a rewrite: once Office documents are converted (#TBD Wave 2), the same dialog offers *original* and *converted PDF* side by side. Design the header for a small set of document actions, not for one button.
- Audit: `knowledge.source_viewed` exists for views. A download is a materially different governance event — "who took the actual spec sheet out of the building" — so it needs to be distinguishable in the audit trail. Follow AGENTS.md § *API Routes And Audit Trail*: `outcome` set, no plaintext PII in `detail`, resource names snapshotted.

## Owns

- `packages/web/src/components/assistant-ui/attachment-preview.tsx` + tests
- `packages/web/src/app/api/agents/[agentId]/workspace-file/route.ts` + tests

**Do not touch:** anything under `lib/knowledge/`, the `pinchy-knowledge` plugin, the reindex UI, `next.config.ts`.

## Depends on

Branch `feat/kb-sources-and-index-ui` (rewrites `PdfDialog` — it gained its own header row instead of controls floating over the browser's PDF viewer). Start after that lands or you will rebase onto a rewritten component.

## Acceptance

- [ ] Download reachable from the viewer, keyboard-accessible, labelled
[...]

PR #1021, opened 2026-07-31
fix: backport 30 user-facing fixes to release/0.9 for v0.9.1

clemenshelm:
Backport round for **v0.9.1**: 30 fixes cherry-picked from `main`, plus the upgrade-notes section without which `pnpm release 0.9.1` refuses to start.

## Selection rule

A fix belongs here if **the bug reaches a v0.9.0 user**. That is narrower than "it's a `fix:` commit": most of what has landed on `main` since the cut repairs features that are not in 0.9.0 at all, and picking those would drag feature code into a patch release.

Deliberately **out**: everything under #139 (Automations / Inbox Agent), #703 (file delivery to chat), #788 (`pinchy_generate_file`), #714 (reindex UI), #803 (Eval v2), the embeddinggemma switch, the Office→PDF converter, `security.txt`, and all `dev` / `ci` / `test` / `eval` scoped commits — none of it exists in v0.9.0, so there is no bug there to fix. Also out: `13fed80b` (#1013, retry-gate), which brings a new migration `0064` and a new route, and whose origin is an intermittent E2E failure rather than a reported user problem — a schema migration would contradict what this branch's own upgrade notes promise.

## What is in

**The credentials endpoint now checks which agent is asking** (#987) — 5 commits. `GET /api/internal/integrations/:connectionId/credentials` authorized on the gateway token alone, and that token is one shared secret inlined into every plugin's config block. Any plugin could name any `connectionId` and receive a decrypted Odoo password or mailbox token belonging to a different agent. Callers now pass `?agentId=`, the grant is checked before anything is decrypted, and a refusal writes an `integration.credentials_denied` audit row. Verified open in v0.9.0: the route there authorizes on `validateGatewayToken(request.headers)` and nothing else.

**Chat survives a backgrounded tab** (#895, #956) — 6 commits. The reconnect loop pauses while hidden, resumes on visible, reconciles history over the socket it already has, and the wait for the initial history frame is bounded. This is the cluster with the widest daily reach: phones, installed PWAs, and any tab left in the background.

**The false localhost security banner** (#929) — 2 commits. Every local install was told to lock a domain it does not have.

**A failed migration says what failed** (#932) — 2 commits. The boot report guessed at the broken file from a count. It now names the target and reports the real error; the boot step moves off the `drizzle-kit` CLI, which swallowed failures, onto drizzle-orm's own migrator.

**Audit-chain verification** (#963) — 2 commits. The verify sweep could step over rows written while it ran, and the checkpoint could stall when the concurrency probe failed. This is the product's integrity guarantee, so a sweep that silently covers less than it claims is worth a patch release on its own.

**OpenClaw config-volume ownership** (#1006) — 2 commits, plus the EACCES that the empty-provider path was swallowing.

Then: settings back-navigation, the mobile keyboard covering the composer, the setup-wizard rename, provider cloud-catalog bounding, the vision probe testing sight instead of HTTP 200, the KB template no longer pointing agents at the file tools, usage-poll log noise, and bug-report labelling.

## What I could not pick cleanly — 8 commits, listed rather than forced

Each conflicts, and each conflict is worth a separate decision rather than a hurried resolution on a release branch:

| Commit | Why it conflicts |
| --- | --- |
| `6f0e61b2` frame-header class | `next.config.ts` collides, and one route it covers does not exist in v0.9.0. Security-relevant — **worth a deliberate pass** |
| `ffe11ece`, `a9d299a4` odoo stop-and-report | Both sit on #926's read-back verification, a feature on `main` |
| `ec77d108`, `357945e5` KB citations | `pinchy-knowledge` has drifted; the fixes assume the locator work |
| `5d789d63` citation hardening | Needs `lib/knowledge/source-links.ts`, which does not exist here |
| `245d9126` HEAD without a body | Sits on the KB source-serving route from #946/#980 |
| `7163982f` late preflight | Sits on #975's migration-target naming |
| `1fd87c0e` scan classifier | `pdf-extract.ts` has drifted |

Odoo is a core v0.9.0 integration, so `ffe11ece`/`a9d299a4` are the ones I would look at next if you want a round three.

## Conflicts, and the one shape they all had

Two of the picks conflicted in a way worth naming, because it recurred: **on a release branch a conflict marker is a place where two unrelated changes touch adjacent lines, not a choice between two versions of one change.**

- `f9ee9476` — the block held two `describe`s: a `knowledge.source_downloaded` test for #934 (not in v0.9.0) and the `integration.credentials_tested` test that is the commit's actual point. Dropping the block wholesale would have landed the fix with its test silently removed — the exact failure the no-test-removal rule exists to catch, arriving through a merge conflict instead of an edit.
- `e040b4c4` — same shape in `audit.ts`: `integration.credentials_denied`, the event this fix adds, sat directly above three `email_workflow.*` events from #139. "Theirs" would have imported audit types for a feature that does not exist here; "ours" would have dropped the fix's own event.

Both were resolved by hand and then verified: comparing each pick's diff against its `main` source, **29 of 30 are byte-identical**, and the one that differs (`989ff547`) differs only by the omitted #934 block.

## Review pass

Reviewed the assembled diff afterwards; two findings, both fixed in `25676bdb`:

- The per-agent-ownership note had been cherry-picked into the **frozen** v0.8.0 → v0.9.0 section, because `main` has no `from v0.9.0` section yet and that is where the commit found the end of the file. On this branch the same fix was already described under 0.9.1, so the file documented it twice and told a v0.9.0 upgrader about behaviour v0.9.0 does not have. (`main` still carries the misplacement — separate change.)
- The KB-template fix had no upgrade note, and needs one: `generateAgentsMd` runs in `POST /api/agents` and nowhere else, so it rewrites nothing on upgrade. Every KB agent created under v0.9.0 keeps the `## File Access` block that tells it to `pinchy_ls` before answering, and keeps producing answers without citation numbers. The note now names the block and where to delete it.

## Gates

Green on this branch: `typecheck` (incl. tests) · 9729 unit tests (698 files) · **324 db tests against a real pgvector Postgres** · 9 plugin packages typecheck · 776 plugin tests · 464 script tests · `format:check` · `lint` (0 errors) · docs build.

`test:db` is the one CI caught last round and I had not run: `db-migrate-runner.integration.test.ts` squatted on `agent_delivered_files` (main's 0058, feature #703, absent here), so every migration applied, the runner exited 0, and a test whose whole point is "a FAILURE is reported" found nothing to report. Re-anchored on `0056_openai_compatible_providers`, the last migration that exists on this branch.

Not run locally: E2E and integration — those need the Docker stack, and CI has them.

## Not merging this myself

Rebase-only per repo convention, and the release scope is your call.


clemenshelm:
## Round two: the credentials-endpoint scoping (#987)

Five more commits, picked after #1019 merged to `main`. This is the one I flagged in the description as the only open candidate with real weight, and it is now in:

- `e040b4c4` fix(security): scope the credentials endpoint to the agent that was granted the connection
- `6f0d7162` test(security): pin WEB_SEARCH_TOOLS to the tools build.ts emits the connection for
- `1b92cc29`, `0acd612e` — the docs that were still stating the token-only claim
- plus an upgrade note, because a privilege escalation across agent boundaries should not have to be inferred from a commit list

The hole is open in v0.9.0: the route authorized on `validateGatewayToken(request.headers)` and nothing else, and that token is a single shared secret inlined into every plugin's config.

### The conflict, and why it needed hand-resolution

`audit.ts` conflicted, and the marker block again mixed two unrelated things: `integration.credentials_denied` — the new audit event that IS this fix — sitting directly above three `email_workflow.*` events from #139, which the 0.9 line does not carry. Taking either side wholesale would have been wrong in a different direction: "theirs" drags in audit events for a feature that does not exist here, "ours" drops the fix's own event and leaves the route writing a row the type does not declare.

Same shape as the `f9ee9476` conflict in round one. Two for two, so it is worth naming as the pattern rather than the coincidence: **on a release branch, a conflict marker is a place where two unrelated changes happen to touch adjacent lines, not a choice between two versions of one change.**

### Deliberately left out

`13fed80b` fix(#1013) — the duplicate-write retry gate. It ships a new migration (`0064_wide_captain_stacy`) and a new `retry-gate` route, and its origin is an intermittent E2E failure rather than a reported user problem. A schema migration would also contradict what the upgrade notes on this branch promise. It belongs in 0.10.

### Gates, re-run in full

9729 unit tests · 324 db tests against a real pgvector Postgres · 9 plugin packages typecheck · 776 plugin tests · 464 script tests · typecheck incl. tests · format · lint 0 errors.

I ran `test:db` this time — that is what CI caught last round and I did not.

clemenshelm:
## Round three: empty, and why

Both candidates I named in the body fail the selection rule on inspection, not just on conflict:

- **`ffe11ece` / `a9d299a4` (odoo stop-and-report).** `verifyCreate` / `verifyWrite` appear **zero** times in v0.9.0's `pinchy-odoo`. The read-back verification from #720/#926 is not in this line at all, so there is no error message to reword and no bug reaching a v0.9.0 user. I had called these the next-best candidates; that was wrong.
- **`6f0e61b2` (frame-header class).** The `workspace-file` route exists here, but nothing in the UI consumes it — only lib-level references. The commit's own message says as much: "the same latent blank-viewer defect as artifacts, minus a shipped consumer". Latent, not user-reaching. The drift guard it adds is infrastructure, which this branch does not take.

So the scope stands at 30 fixes.

## The integration shard that went red

`Integration Tests (OpenClaw + Fake Ollama) (1/2)` failed on `25676bdb`: 12 of 23 tests, every one of them downstream of `errorCode=INVALID_REQUEST ... unknown agent id "04b53f47-…"` — the agent never entered the OpenClaw runtime after `[integration-setup] Restarting OpenClaw container to reload config...`. That is the #200 reload race, not a broken assertion.

Classified as transient rather than assumed to be: the **same job passed on `eb633335`**, which already contains every code change in this PR. The delta between the green run and the red one is `git diff --stat eb633335 25676bdb` → one docs file and one blank line in a vitest type test. Neither is reachable from the integration stack. Identical runtime code, two greens and one red.

## Two more docs fixes (`25676bdb`, `a4318e11`)

Reviewing the assembled diff turned up upgrade notes filed under the wrong release — and then a third instance when I checked the rest of the file the same way. The method: compare every `## Upgrading from vA to vB` section against `git show vB:docs/…/upgrading.mdx`. **4 of 7 released sections had drifted from their own tag**, three by gaining content after the tag.

- The per-agent-ownership note sat in the frozen v0.8.0 → v0.9.0 section (`main` has no `from v0.9.0` section yet, so the cherry-pick appended there), duplicating what 0.9.1 already said.
- `d40bb678` had put **two** things in the v0.7.0 → v0.8.0 section after v0.8.0 shipped: the invite-history note, and a `0044` migration bullet. v0.8.0's journal ends at `0043` — that section named a migration the release does not contain. The dangling "(see the note below)" was the seam. v0.8.0 is now byte-identical to its tag again.
- The KB-template fix needed a note it did not have: `generateAgentsMd` runs only in `POST /api/agents`, so every KB agent created under v0.9.0 keeps the `## File Access` block and keeps answering without citations. The note says which block to delete and where.

`main` carries the same two misfilings and needs the identical moves.

PR #1041, opened 2026-08-04
fix(deps): clear the 20 advisories published against the existing lockfiles

clemenshelm:
## Why

`osv-scan` re-evaluates the **existing** lockfiles against the OSV database on every run, so a batch of advisories published overnight turned `main` and every open PR red at once — with no code change anywhere. #1038 is one of the casualties: it touches seven TypeScript files and no lockfile.

20 findings across 10 packages, all transitive.

## What

Nothing new is added. Each override moves a line that was already pinned, to the version OSV names as fixed:

| package | before | after | advisories |
|---|---|---|---|
| `undici` | 7.28.0 / 8.5.0 | 7.29.0 / 8.9.0 | 5 per major |
| `ip-address` | 10.1.0, 10.2.0 | 10.4.0 | 4 |
| `brace-expansion` | 1.1.16, 2.1.2, 5.0.8 | 1.1.18, 2.1.4, 5.0.9 | 3 |
| `hono` | 4.12.31 | 4.13.0 | 1 |
| `fast-uri` | 4.1.1 | 4.1.2 | 1 |
| `postcss` (docs) | 8.5.22 | 8.5.25 | 1 |

Two details worth keeping:

- **Each `undici` major stays capped below the next one.** Unbounded, pnpm dedupes jsdom's 7.x copy up to the 8.x production copy, and jsdom imports an undici-internal path that 8.x removed — 448 workers die at once. That is PR #554's failure, and the cap is what prevents it.
- **`fast-uri` needed the nested selector `ajv>fast-uri`.** `ajv@8.20.0` pins it *exactly* at 4.1.1, and a version-range override does not rewrite an exact pin. The range form installed cleanly and changed nothing — which is the quiet way this fix fails, so it was verified by re-reading the lockfile rather than by the install exiting 0.

The three residual `astro` advisories stay allow-listed (fixed only in Astro 7, which Starlight 0.38 blocks — #851).

## Verification

- `osv-scanner scan source --config=osv-scanner.toml` over both lockfiles, i.e. the config CI uses → **No issues found**
- `pnpm test` — 775 files, 10912 tests green (this is also the jsdom/undici canary)
- `pnpm test:scripts` — 894 green
- `pnpm typecheck:plugins` — 9 packages clean; web typecheck incl. tests clean
- `eslint .` — 0 errors; `prettier --check .` green
- `pnpm -C docs build` — 67 pages

Docs-not-needed: dependency versions only, no reader-facing behaviour.


clemenshelm:
Superseded by 3b7278ddb, which landed on main while this was in CI and carries the identical set of overrides — same versions, same `ajv>fast-uri` nested selector for ajv's exact pin, same per-major undici caps. Closing rather than reconciling a lockfile conflict against an equivalent fix. #1038 rebases onto main instead.

PR #1107, opened 2026-08-04
fix(security): sign the development license with a key production rejects (#1083)

### LINKED ISSUE #1083: security: committed enterprise license JWT unlocks paid features on any install (unrevocable without code change)
## Finding (2026-08-04 repo review, security pass, P3)

`packages/web/src/app/api/dev/enterprise-toggle/route.ts:8-9` embeds a valid signed license JWT (exp 2088), also present in `docker-compose.dev.yml:20`. The dev route is NODE_ENV-gated, but any self-hoster can paste the committed token into `PUT /api/enterprise/key` and unlock paid features. Revenue/licensing bypass, not a user-security issue — and unrevocable without a code change.

## Proposed approach (needs a product decision, hence issue not PR)

Options: (a) accept and document (dev keys are findable in any AGPL repo history anyway); (b) mint a dev license that the production license validator rejects (e.g. a dedicated dev keypair, prod validator pins the prod public key — dev compose/route use the dev key); (c) short-lived dev tokens generated at stack boot. (b) is the structural fix; history rewrite is not worth it, so rotation of the validating keypair is what actually revokes the committed token.

clemenshelm:
Closes #1083.

## The finding

`packages/web/src/app/api/dev/enterprise-toggle/route.ts` and `docker-compose.dev.yml` shipped the **same license token signed with the production key**, valid until **2036-03-11** (the issue said exp 2088 — that is the epoch value `2088845234`, which is 2036). Only the route that installs it was `NODE_ENV`-gated; the string itself was a working enterprise key, so `PUT /api/enterprise/key` accepted it on any install.

Nothing was ever red about it. A valid key makes every test pass — that is exactly why it survived.

## What this does

**1. A development keypair.** The dev license is now signed by a key that `developmentLicensesHonoured()` consults only outside production. Committing it is safe for that reason. The private half is deliberately **not** committed: nothing at runtime needs it, and a private key in the tree is the shape of the bug being replaced. Regeneration is documented in `license-keys.ts`; the token runs to 2126.

**2. The revocation.** Rotating the production keypair would have invalidated **every customer license**, so instead the production key never honours `DEV_LICENSE_SUBJECT` (`"pinchy-dev"`) — on the valid and the expired path alike. That rule, not a rotation, is what kills a token that is in every clone and every fork and cannot be un-published. A test pins the string, because changing it un-revokes the token.

Verified by reproduction rather than by reading: through the real validator the old token reads `active=false`, and `active=true` with the rule forced off.

**3. A refused key fails closed** (second commit — found by the `review-docs` pass). Returning `INACTIVE` derived the *community* state, where `effectiveVisibility` maps `restricted` → `all`. An install running the leaked key would have **published every agent somebody had deliberately restricted** on upgrade. An authentic-but-refused signature now returns `expired: true` with no claims: management locks, restrictions stay enforced, and the banner degrades to its dateless copy instead of printing "Your license period ended on 11 March 2036".

**4. The tripwire.** `committed-license-tokens.test.ts` fails if any JWT in the tracked tree verifies against the production key, and names the file. It asks jose directly rather than going through `validateLicense`, so it fires even on a token our own policy would reject — the question is whether *we signed it*, not whether we would grant it. It also pins the `docker-compose.dev.yml` copy against `DEV_LICENSE_TOKEN`.

## Residual, stated rather than hidden

A self-hoster can still run their install outside production mode. That degrades the app visibly and is a code-level change in spirit — and under AGPL a determined operator can always patch the validator. The licence is a control, not a boundary. What changes here is that **pasting a string no longer works**, which was the whole of the exposure.

## Also

`AGENTS.md` § Commands claimed `PINCHY_ENTERPRISE_KEY=dev-enterprise` was how you get a dev enterprise key. It is not a valid token, so that command turned enterprise *off* — the dev stack has shipped a license by default all along.

## Verification

Canaried in both directions rather than reasoned about:

- re-introducing the old token into a tracked file → tree guard red, naming the file
- forcing `developmentLicensesHonoured()` to `true` → "unlocks nothing in production" red
- returning `INACTIVE` instead of `REFUSED` → `expected 'community' to be 'expired'`

Gates: `pnpm test` (776 files, 10953 tests), `pnpm test:scripts` (913), `typecheck`, `lint` (0 errors), `format:check`, `pnpm build`, docs build + `check:anchors` + `check:rendered`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -476,13 +476,23 @@ jobs:
       - name: Verify private key matches public key
         env:
           PINCHY_LICENSE_PRIVATE_KEY: ${{ secrets.PINCHY_LICENSE_PRIVATE_KEY }}
+          # "true" on a fork PR, "false" on a same-repo PR, empty on push and
+          # merge_group. Only the first case legitimately has no secrets.
+          FORK_PR: ${{ github.event.pull_request.head.repo.fork }}
         run: |
-          # Fork PRs get no secrets; a missing key must read as "cannot verify
-          # here", not as a red check on every outside contribution. Runs with
-          # the secret on same-repo PRs, merge queue and main.
+          # A fork PR carries no secrets by design, so "cannot verify here" must
+          # not read as a red check on every outside contribution. Anywhere else
+          # the secret is supposed to be present, and its absence means this
+          # check has been silently switched off — which is the one outcome a
+          # security check may never report as green. Skip on a fork, fail
+          # otherwise.
           if [ -z "$PINCHY_LICENSE_PRIVATE_KEY" ]; then
-            echo "::notice::PINCHY_LICENSE_PRIVATE_KEY not available (fork PR or unset secret) — skipping keypair verification."
-            exit 0
+            if [ "$FORK_PR" = "true" ]; then
+              echo "::notice::Fork PR carries no secrets — keypair verification cannot run here."
+              exit 0
+            fi
+            echo "::error::PINCHY_LICENSE_PRIVATE_KEY is unset on a run that should have it (event=${GITHUB_EVENT_NAME}). The keypair check cannot verify anything; restore the repository secret rather than letting this pass."
+            exit 1
           fi
           npm install jose
           node -e "
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -477,19 +477,31 @@ jobs:
         env:
           PINCHY_LICENSE_PRIVATE_KEY: ${{ secrets.PINCHY_LICENSE_PRIVATE_KEY }}
         run: |
+          # Fork PRs get no secrets; a missing key must read as "cannot verify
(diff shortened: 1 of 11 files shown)


clemenshelm:
Der `Verify License Keypair`-Job las den Public Key noch aus `enterprise.ts`, obwohl dieser Branch die Keys nach `license-keys.ts` verschoben hat — der Check konnte auf seinem eigenen PR nie grün werden. Gefixt in 66601c04: liest jetzt `license-keys.ts`, ankert auf `PRODUCTION_PUBLIC_KEY` (die Datei enthält auch den Dev-Key; ein First-Match-Regex würde sonst ggf. gegen den falschen verifizieren) und skippt mit Notice, wenn `PINCHY_LICENSE_PRIVATE_KEY` fehlt (Fork-PRs bekommen keine Secrets). Branch dabei auf origin/main rebased.

PR #1109, opened 2026-08-04
fix(plugins): one credential client, and stop reading "401" out of prose

### LINKED ISSUE #1077: Plugins: unify the 4x-copied credential/HTTP client (drifted auth matchers, cache keys, non-idempotent retry)
## Finding (2026-08-04 repo review, plugin pass)

`fetchCredentials` + SecretRef shape assertion + `reportAuthFailure` + 5-min TTL cache + `withAuthRetry` exist in near-copies in pinchy-odoo (`index.ts:190-225, 2436-2491`), pinchy-email (`index.ts:627-684, 734-885`) and pinchy-web (`index.ts:71-197`), and they have already drifted:

- **Auth-error detection differs per plugin**: odoo matches `'access denied'|'invalid api key'|'401'|'authenticat'`, email `'401'|'invalid credentials'|'invalid_grant'|'token has been expired'|'unauthorized'`, web `'401'|'unauthor'|'invalid api'`. Substring '401' matches any error that happens to contain 401 (an amount, a record id) and then `withAuthRetry` re-executes a NON-idempotent closure (odoo_create ends in `client.create`).
- **Cache keys differ**: email keys `agentId:connectionId`, odoo only `agentId` — after a connection hot-swap the old odoo client can live up to 5 minutes.

## Proposed approach

Either a shared `credential-client.ts` copied per plugin with a textual drift guard (the `normalize-docx-table-html` pattern), or at minimum: unify the auth matchers + cache keys, and move non-idempotent operations out of retry closures. Timeouts were fixed separately in PR #1048.

clemenshelm:
Closes #1077.

## What was wrong

`pinchy-odoo`, `pinchy-email` and `pinchy-web` each carried their own copy of fetch-credentials / cache / classify-auth-error / retry-once, and the three copies had drifted — three cache-key conventions, three auth matchers, and in all three a test for the bare substring `401`:

```
odoo   'access denied' | 'invalid api key' | '401' | 'authenticat'
email  '401' | 'invalid credentials' | 'invalid_grant' | 'token has been expired' | 'unauthorized'
web    '401' | 'unauthor' | 'invalid api'
```

Three digits are not a status. Odoo's `MissingError` names the record (`Records: sale.order(401,)`), an unbalanced-entry error names the amount, Graph echoes a recipient that may be a 401(k) plan, and Microsoft stamps a GUID request-id on every error body — roughly one in seventy contains `401` by chance.

The cost is not a wasted retry. The second such failure POSTs `report-auth-failure`, which flips the connection to `auth_failed`, shows admins a "reconnect" banner in Settings → Integrations, and writes an `integration.auth_failed` audit row for an auth failure that never happened. (Credentials keep being served — `resolve-credentials.ts` only refuses `pending` — so this is a false alarm and audit-trail pollution, not an outage.)

## What changed

`credential-client.ts` is the single implementation, duplicated byte-for-byte per plugin. The duplication is forced: each plugin directory is mounted into the OpenClaw container standalone (`./packages/plugins/<name>:/root/.openclaw/extensions/<name>`), so an import escaping the plugin directory does not resolve at runtime. Same shape as `normalizeTableHtml`, same answer — duplicate the source, guard the copies.

- **`isAuthError`** reads a structured status first (`status` / `statusCode` / `response.status` / numeric `code`) and otherwise matches words, never digits.
- **`BraveSearchError` and `GraphRequestError` now carry `status`.** This is the price of the narrowing and paying it is part of the change: Brave's real 401 body is `{"code":"SUBSCRIPTION_TOKEN_INVALID"}` and contains no auth word at all, so without the status a stale key would stop triggering a refresh.
- **`credentialCacheKey(agentId, connectionId)`** — odoo keyed on the agent alone. Honest caveat: reachability today depends on OpenClaw re-running `register()` on a config change, which rebuilds the cache anyway. The key is still wrong, and "the other input happens to be constant" is a property of today's config, not of the cache.
- **`trackMutations`** refuses a retry once a mutating call has **resolved**. A rejected write changed nothing (an Odoo RPC is transactional, a rejected send did not send), so retrying it under a fresh token is exactly what the retry is for — marking on entry would disable the refresh for every write tool. No closure runs a step after its mutating call today, so this is a **tripwire**, not a live fix.
- **Drift guard** `plugin-credential-client-drift.test.ts` compares the copies byte-for-byte and also fails on a plugin that grows a second classifier in its own `index.ts`.

## Verification, by reproduction

| canary | result |
| --- | --- |
| re-add `lower.includes("401")` to the classifier | 7 red (5 classifier cases, the odoo read, the odoo create) |
| restore the `agentId`-only cache key | 2 red |
| append one line to any copy of `credential-client.ts` | drift guard red |

Gates run locally: `pnpm test` (776 files / 10 967 tests), `pnpm test:scripts` (913), `pnpm test:plugins`, `pnpm typecheck:plugins`, `pnpm -C packages/web typecheck`, `pnpm -C packages/web lint` (0 errors), `pnpm format:check`.

## Notes for review

- The module's unit tests live once (`pinchy-odoo/__tests__/credential-client.test.ts`) and count as coverage for all three copies only for as long as the drift guard passes. That argument is written into the guard's docblock.
- The issue title says "4x-copied"; there are three copies — no other plugin calls the credentials route.
- No user-visible surface moved, so no docs change. `AGENTS.md` gains a "Three Digits Are Not A Status" section, as every guard here does.

Docs-not-needed: internal plugin behaviour, no reader-facing surface changed


clemenshelm:
Rebased onto main after `e74f7628 fix(plugins): bound every plugin fetch call against a hung endpoint` landed — it conflicted with this PR in all three `index.ts` files, and the conflict was substantive rather than textual.

That commit added `AbortSignal.timeout(10_000)` to exactly the two fetches this PR moves into `credential-client.ts` (credentials fetch, report-auth-failure). Resolving in favour of this branch would have silently reverted the hardening. So:

- `credential-client.ts` carries `FETCH_TIMEOUT_MS = 10_000` and both calls pass `AbortSignal.timeout(...)`, matching what each plugin used before.
- The now-dead `FETCH_TIMEOUT_MS` constant was removed from each `index.ts` and replaced with a comment saying where the bound moved. A constant that no longer bounds anything is worse than no constant.
- `isAuthError` deliberately does not classify an `AbortError`, so a hung Pinchy container can never flag a connection `auth_failed`.

The check that this is preserved rather than asserted: main's own test — "aborts a hung credentials fetch via AbortSignal.timeout(10_000) instead of blocking the tool call forever" — passes unchanged against the shared implementation, as do the Graph and Brave equivalents.

Re-ran after the rebase: `pnpm test` (778 files / 11 050 tests), `test:scripts` (913), `test:plugins` (9 packages), `typecheck:plugins`, `-C packages/web typecheck`, `format:check`.

Issue #721, opened 2026-07-14
feat(pinchy-odoo): deterministic duplicate guard — block vendor-bill creates that match an existing bill


clemenshelm:
## Why

Eval-v1 (#669, published at https://heypinchy.com/reliability) ran a **duplicate-guard scenario**: the vendor bill the agent is asked to file already exists in the ERP (posted, same vendor, same invoice reference). Result: **13 of 14 models filed the duplicate anyway**. Of the 96 failing runs with complete trajectories, 90 never queried the ERP before writing, and 6 queried, saw the existing bill, and filed anyway.

Like the zero-read-back finding, this is behavioral and capability-independent: models are trained to execute the instruction, not to check state first. In production this is double payment — one of the failure modes enterprises fear most from autonomous agents.

## Proposal

Enforce the check deterministically in the `pinchy-odoo` plugin: before an `odoo_create` on `account.move` (vendor bills first: `move_type = in_invoice`), the plugin searches for an existing bill with the same partner + reference (`ref` / vendor invoice number).

- On hit: **block the create** and return a structured result naming the existing bill (`{ id, name, state, amount_total, date }`) so the model can relay it to the user instead of writing.
- Provide an explicit override parameter (e.g. `allow_duplicate: true`) for legitimate re-filings; overridden creates get their own audit detail so the decision is traceable.
- Audit the blocked attempt (`outcome: failure`or a dedicated blocked outcome detail) — a blocked duplicate is exactly what the audit trail should show.

This is the idempotency pattern applied at the only place it can live for ERP integrations: Odoo's API has no native idempotency keys for vendor bills, and generic guardrail frameworks validate call *shape*, not state *effects*. Pinchy owns the tool layer, so Pinchy can own the guarantee.

## Validation via the eval harness (keep-or-revert)

- Re-run the **duplicate-guard** scenario through the guarded plugin: expected outcome ~14/14 pass (no duplicate created, user informed of the existing bill).
- Re-run **happy-path** to rule out false positives (guard must not block a legitimate first-time bill).
- Watch the trajectories for models that fight the block (retry loops against the structured refusal) — the hard-rejection scenario showed some models loop on refusals; the block message must be phrased so models relay it rather than retry.
- If results do not improve, or the guard causes loops/regressions elsewhere, revert.

## Acceptance

- [ ] TDD: failing plugin unit tests first (hit → block with existing-bill snapshot, miss → create proceeds, override path, permission interaction)
[...]

Issue #720, opened 2026-07-14
feat(pinchy-odoo): plugin-side write verification — read back every create before reporting success


clemenshelm:
## Why

Eval-v1 (#669, published at https://heypinchy.com/reliability) produced a flagship finding: in the **silent-failure scenario** (the mock ERP returns a plausible-looking id but persists nothing), **0 of 162 graded runs across all 14 models performed a verifying read after the write**. Every model trusted the bare tool return value and reported success to the user while the ERP stayed empty.

This is a *behavioral* failure, not a capability failure — the strongest open-weight models (kimi-k2, deepseek, glm) fail it exactly like the weakest. Waiting for better models will not fix it, and prompt-only mitigations depend on model compliance (see #385: the Step-7 prompt mandate is skippable by any non-compliant model — and the eval shows that unprompted, literally no model reads back).

## Proposal

Move write verification into the tool layer, where it is deterministic: after every `odoo_create` (and value-bearing `odoo_write`) on transactional models, the `pinchy-odoo` plugin itself reads the record back and compares core fields before returning to the model.

- On match: return the normal success result, enriched with the verified values, and audit the tool call with `verified: true`.
- On mismatch or missing record: return a structured error ("write reported id X but the record could not be verified") so the model must surface the failure instead of narrating success; audit with `outcome: failure` and `verified: false`.
- Scope: start with `account.move` (+ its line models), the highest-stakes writes; generalize if the eval supports it.

This turns a universal model weakness into a platform guarantee: the customer does not have to hope their model reads back, because Pinchy does it deterministically.

## Relationship to #385

#385 proposes *detecting* whether the model performed the verifying read (audit scan / in-session reminder / hard rejection). This issue makes the question moot for the covered models/operations: the plugin performs the read itself. #385's option 1 (retrospective audit query) stays useful as coverage monitoring; options 2/3 are subsumed for the models covered here.

## Validation via the eval harness (keep-or-revert)

The Eval-v1 harness is the acceptance test. After implementation:

- Re-run the **silent-failure** scenario through the governed tool layer: expected outcome is that all models now *report the failure* (the injected silent no-op becomes a hard, visible error), i.e. false-success drops to zero by construction.
- Re-run **happy-path** and **line-items** to check for regressions (added latency, token overhead, or the enriched result confusing weaker models).
- If the guarded runs show no improvement or degrade other scenarios, revert — the harness gives us that option cheaply.

## Acceptance

[...]

PR #1114, opened 2026-08-04
fix(plugins): unify drifted credential-client copies across odoo/email/web

### LINKED ISSUE #1077: Plugins: unify the 4x-copied credential/HTTP client (drifted auth matchers, cache keys, non-idempotent retry)

clemenshelm:
## Summary

- Extracts the credential-fetch / auth-retry building blocks (`isAuthError`, `reportAuthFailure`, the 5-min credential/client cache, `withAuthRetry`'s matcher) that were near-copied across `pinchy-odoo`, `pinchy-email`, and `pinchy-web` into a single `credential-client.ts`, copied byte-identically into each plugin directory (bundle isolation — same rationale as `normalizeTableHtml` / docx-table-html across `pinchy-files` and the web composer). Pinned by a new textual drift guard: `packages/web/src/__tests__/lib/credential-client-drift.test.ts`.
- Unifies the auth-error matcher (`isAuthError`) to the union of all three plugins' previously-independent phrase lists, and makes the `401` check precise: it now requires a real HTTP-401 envelope (`"Graph 401:"`, `"(401)"`, `"HTTP 401"`, `"401 Unauthorized"`, `"status code 401"`) instead of a bare substring match. Before this, a benign amount or record id containing the digits `401` (e.g. an invoice total of `401.00`, a record id of `401`) could be misclassified as an auth failure, causing `withAuthRetry` to invalidate the cached client/adapter and **replay the entire closure — including a non-idempotent write** (`odoo_create`'s `client.create`, `email_send`'s `adapter.send`) — a second time.
- Unifies the credential/client cache key to `agentId:connectionId` everywhere a per-agent cache exists. `pinchy-odoo` previously keyed on `agentId` alone, so after a connection hot-swap the old client could stay cached for up to 5 minutes. `pinchy-email` already used `agentId:connectionId`; `pinchy-web`'s single-entry cache is deliberately left un-keyed (documented in the code) because the Brave key is instance-wide by design — that is not a drift, it's a different, intentional cache shape.

`fetchCredentials()` itself stays per-plugin: each plugin fetches a differently-shaped credential object (Odoo: `{url, db, uid, apiKey}`; email: OAuth-or-IMAP dispatch on a `type` field; Brave: `{apiKey}`) with its own SecretRef-shape assertion (#209), and unifying that is out of scope.

Closes #1077

## `withAuthRetry` callsite audit (non-idempotent writes)

Per the issue's request to check every `withAuthRetry` callsite for replayed non-idempotent operations, here's what I found:

- **`pinchy-odoo`**: the only genuinely non-idempotent operations reachable through `withAuthRetry` are the three `client.create(...)` calls (`odoo_create`, the `mail.activity` create inside `odoo_schedule_activity`, and `ir.attachment` create inside `odoo_attach_file`). In every case, `client.create(...)` is the terminal statement of its closure — nothing else runs afterward that could produce a second false-positive-triggered replay within the same attempt. Everything else wrapped in `withAuthRetry` (`searchRead`, `write`, `unlink`, `callMethod` for confirm/reconcile/undo-reconciliation) is either naturally idempotent (a `write` with the same values) or fails loudly on a second application (Odoo state-transition guards reject an already-confirmed/already-reconciled record) rather than silently duplicating.
- **`pinchy-email`**: `email_send` (`adapter.send`) and `email_draft` (`adapter.draft`) are the non-idempotent writes; both are single-call closures.
- **`pinchy-web`**: no non-idempotent writes reachable through `withAuthRetry` (search/fetch are read-only).

I did not restructure these closures to physically separate "credential refresh" from "the write" — that would be a materially larger, riskier change touching heavily-commented, already-tested write-verification logic (`#720`/`#721` duplicate-guard + read-back verification in `odoo_create`) for marginal additional safety. Given that a real Odoo/OAuth auth failure is rejected at the RPC/HTTP layer *before* the write executes server-side, the precise-matcher fix removes the practical risk this issue raises: a false-positive can no longer be triggered by ordinary response data, so the replay path is (in practice) only ever taken for a genuine, not-yet-persisted auth failure. Flagging this here as an explicit, deliberate scope boundary rather than silently declaring the callsite audit "done."

## Test plan

- [x] New drift guard: `packages/web/src/__tests__/lib/credential-client-drift.test.ts` — 3 tests, pins all three `credential-client.ts` copies to be byte-identical modulo comments/whitespace.
- [x] New behavior tests (duplicated per plugin, same rationale as the source duplication — each package's own `pnpm test:plugins` run exercises its own copy): `credential-client.test.ts` in `pinchy-odoo/__tests__/`, `pinchy-email/__tests__/`, and `pinchy-web/` (29 tests each) — cover `isAuthError`'s legitimate phrases, the precise HTTP-401 envelopes, the explicit "bare 401 in unrelated data must NOT match" cases from the issue, `credentialCacheKey`'s hot-swap-safety property, and `reportAuthFailure`'s POST shape/truncation/best-effort-swallow behavior.
- [x] All pre-existing plugin tests (`pinchy-odoo` 482, `pinchy-email` 396, `pinchy-web` 129) still pass unchanged against the new matcher — confirms no regression in real-world auth-failure detection (e.g. `"Access Denied: invalid api key"`, `"HTTP 401: Invalid Credentials"`, `"401 Unauthorized"` all still match via their textual substrings independent of the tightened 401-envelope check).

### Gates (from repo root)

- `pnpm typecheck:plugins` — all 9 plugin packages typecheck cleanly.
- `pnpm -C packages/web` test-file typecheck (`tsc --noEmit -p tsconfig.typecheck.json`) — clean.
- `pnpm test:plugins` — pinchy-odoo 482/482, pinchy-email 396/396, pinchy-web 129/129, all other plugins unchanged and passing.
- `pnpm test:related` — 661/661 passed.
- `pnpm format:check` — clean.
- `pnpm test` (full suite) — 780 test files passed / 4 skipped, 11090 tests passed / 18 skipped.

`Docs-not-needed: internal plugin refactor, no user-facing change` trailer is on the commit (no product-behavior or API-shape change).

### DIFF
--- a/packages/plugins/pinchy-email/__tests__/tools.test.ts
+++ b/packages/plugins/pinchy-email/__tests__/tools.test.ts
@@ -1406,6 +1406,49 @@ describe("email_send", () => {
     });
   });
 
+  it("does not re-send when the failure merely contains the digits 401 (pinchy#1077)", async () => {
+    // The pre-#1077 matcher tested `msg.includes("401")` over the whole
+    // error message, so any provider error carrying those digits — a quota
+    // figure, a size, a message id — was classified as an auth failure.
+    // withAuthRetry then invalidated the adapter and replayed the entire
+    // closure, and for email_send that closure IS the send: the recipient
+    // gets the mail twice.
+    //
+    // Asserted on the tool, not on isAuthError: the unit test proves the
+    // matcher rejects the string, this proves the send is not repeated.
+    mockSend.mockRejectedValue(new Error("Recipient mailbox is 401 MB over quota"));
+
+    const configWithSend: PluginConfig = {
+      ...testConfig,
+      agents: {
+        "agent-1": {
+          connectionId: "conn-1",
+          permissions: { email: ["read", "draft", "send"] },
+        },
+      },
+    };
+    const tools = createApi(configWithSend);
+    const tool = findTool(tools, "email_send", agentId)!;
+
+    const result = await tool.execute("call-1", {
+      to: "recipient@test.com",
+      subject: "Sent Subject",
+      body: "Sent body text",
+    });
+
+    // Read the count, THEN reset, THEN assert. `vi.clearAllMocks()` in
+    // beforeEach clears recorded calls but keeps a standing implementation
+    // (and any unconsumed `…Once`), so a rejection left behind here would
+    // surface as a failure in whichever later test calls send without its
+    // own setup. Resetting before the assertions means a red assertion
+    // cannot skip the cleanup and turn one failure into several.
+    const sendCalls = mockSend.mock.calls.length;
+    mockSend.mockReset();
+
+    expect(result.isError).toBe(true);
+    expect(sendCalls).toBe(1);
+  });
+
   it("reports a direct send honestly when the adapter returns messageId: null (Graph's 202-with-no-location case) instead of fabricating an empty id", async () => {
     mockSend.mockResolvedValue({ messageId: null });
 
--- a/packages/plugins/pinchy-odoo/__tests__/tools.test.ts
+++ b/packages/plugins/pinchy-odoo/__tests__/tools.test.ts
@@ -3719,6 +3719,83 @@ describe("client caching (#209 layer 2: credentials fetched lazily, cached)", ()
     expect(mockSearchRead).toHaveBeenCalledTimes(2);
   });
 
+  it("refetches after a connection hot-swap instead of serving the previous connection's client (pinchy#1077)", async () => {
+    mockSearchRead.mockResolvedValue({
+      records: [],
+      total: 0,
+      limit: 100,
+      offset: 0,
+    });
+
+    // A hot swap re-points the SAME agent at a different integration
+    // connection; the agentId does not change. Keying the client cache on
+    // agentId alone therefore kept the OLD connection's OdooClient — and
+    // with it the old Odoo instance's credentials — serving tool calls for
+    // up to CREDENTIAL_CACHE_TTL_MS after the swap.
+    //
+    // This asserts the behaviour, not the helper: credentialCacheKey's own
+    // unit test can only prove that two strings differ, which stays true
+    // even if index.ts goes back to caching on `agentId`.
+    const agentConfigs: Record<string, AgentOdooConfig> = {
+      [agentId]: { ...agentConfig, connectionId: "conn-before-swap" },
+    };
+    const tools = createApi(agentConfigs);
+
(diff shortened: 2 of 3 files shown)


clemenshelm:
Rebased onto main — and reduced, because main solved most of this in the meantime.

`f87aed9a fix(plugins): one credential client, and stop reading "401" out of prose` landed the same unification (#1077) from the other direction: one `credential-client.ts`, byte-duplicated per plugin, keyed on `agentId:connectionId`, with an envelope-based `isAuthError`. That is this PR's first commit, so it is dropped rather than merged — seven conflicts of two implementations of the same thing is not a merge worth doing.

What main does **not** have, and what this PR is now:

1. **Regression tests at the tool boundary.** Main's coverage is helper-level (`credentialCacheKey("a","x") !== credentialCacheKey("a","y")`, `isAuthError` rejects `"401.00"`). Both stay green if `index.ts` goes back to `cache.get(agentId)` and `msg.includes("401")` — so neither bug #1077 describes had a regression test. Verified by canary: reverting both fixes turns these red (plus three of main's own), restoring them turns all 434 green.
2. **The drift guard discovers its copies.** It listed `COPY_PLUGINS = ["pinchy-email", "pinchy-web"]`, so a fourth copy would drift unwatched while the guard stayed green. Now derived from the filesystem. Canary: a fourth, one-line-different copy in `pinchy-docs` fails it by name.

PR #1124, opened 2026-08-05
fix(security): give every API key a request budget on /api/v1/* (#1086)

### LINKED ISSUE #1086: security: /api/v1/* API keys are unthrottled (documented gap) — add a limiter
## Finding (2026-08-04 repo review, security pass, P3 — recording the known gap so it stays visible)

`src/lib/auth.ts:249-253` disables Better Auth's rate limiting for the apiKey plugin with a comment that says it is 'a gap, not a delegation': `/api/v1/*` requests authenticated by API key have no throttle at all. `claimScopeDenialSlot` bounds the audit noise of scope denials, but nothing bounds request volume or brute-force attempts against key values.

## Proposed approach

Per-key (and per-IP for invalid-key attempts) FixedWindowRateLimiter in `withApiKey` (`lib/api-auth.ts`) — the same in-process pattern now used by the Telegram pairing route (PR #1049) and password change (PR #1054). Include a 429 with Retry-After; audit throttled bursts through the existing scope-denial slot mechanism.

clemenshelm:
Closes #1086.

## The gap

`/api/v1/*` was bounded by nothing. Better Auth's apiKey limiter is off — its 10-requests-per-24h default would throttle a real CI pipeline — and its general-purpose limiter guards the auth router, which this surface never enters: `withApiKey` reaches `verifyApiKey` through `auth.api.*`. So an authenticated key could call as fast as the socket allowed, and every made-up string reached a hash plus a database lookup.

## What this does

**A per-key budget** — `claimApiKeyRequest` charges 300 requests/minute against the **verified** key id. That key is the design, not an implementation detail: the map grows with issued keys rather than with request volume, and a key id is a fact rather than a spoofable header — the trap `claimScopeDenialSlot` already documents next door. Over budget the wrapper answers `429` with `Retry-After`, rounded up so a client that honours it doesn't land back inside the same window.

Two placement decisions carry weight:

- **Charged before the scope check.** A key probing for scopes it doesn't hold is precisely the caller worth bounding. The denial *row* was already throttled; the requests minting it were not.
- **Charged only after verification.** A 401 has no id to bill, so an unauthenticated flood can't spend a real key's budget.

**A shape check before verification** — every key Pinchy has minted starts with `pinchy_` (`/api/settings/api-keys` is the only creation path, and it passes no prefix). A candidate carrying anything else can't verify, and now says so before the round trip. The prefix became one constant that `auth.ts` imports as `defaultPrefix`, rather than a second string that agrees — drift there would reject every valid key at once.

**An audit row** — `auth.rate_limited`, one per key per window, with `suppressedSinceLastEntry` carrying the volume it stands in for. A row per rejection would make the limiter the log-flooding amplifier it exists to prevent.

## What this deliberately does not do

A **well-formed but wrong** key still costs a lookup, and nothing bounds how many arrive. Bounding that needs an identity for an unverified caller, and the only one available here is a source address the caller sets in a header. Keys are 64 random characters, so guessing one isn't the threat — load is, and a header-keyed limiter bounds no load while allocating an entry per value an attacker invents. That limit belongs on the reverse proxy, where the connection is.

This is stated rather than implied: the reference has its own "What this does not cover" heading, and the three places that claimed the surface was unthrottled (the reference, `auth.ts`, the 401 test's rationale) are corrected rather than left to rot.

## Tests

TDD throughout. The window arithmetic is pinned deterministically with an injected `now` (`api-key-rate-limit.test.ts`); the translation into a 429, a header and a row is pinned separately (`with-api-key.test.ts`), including one test that drives 300 real requests through the wrapper — a wrapper that consulted the limiter but never charged a *successful* request would pass every seeded test while throttling nothing.

## Gates

`pnpm test` 10969 ✓ · `pnpm test:scripts` 913 ✓ · `typecheck` ✓ · `lint` 0 errors · `format:check` ✓ · `pnpm build` ✓ · docs build + `check:anchors` + `check:rendered` + docs tests ✓


clemenshelm:
Superseded by 8804fee7, which landed on main and closed #1086 — with a per-IP invalid-attempt limiter on top of the per-key budget. That's the piece I argued against building here, and the objection I raised (an unbounded map keyed on a caller-supplied value) is answered there by `INVALID_API_KEY_IP_MAX_TRACKED` failing closed. Theirs is the better implementation; rebasing this would only stand a second limiter next to it.

The one thing that didn't land: `verifyApiKey` runs *before* the invalid-attempt limiter, so that limiter bounds the answer and not the work — a sprayed `sk-live-…` still buys a hash and a database lookup. Carried over as #1151, which is just the prefix shape check against current main.

PR #1161, opened 2026-08-05
fix(security): a download link serves the file it was given, not the name

### LINKED ISSUE #903: Delivered-file grants serve whatever bytes currently sit behind the filename — overwrite & zone-shadowing can swap content under a live grant (shared agents)
## Context

The #703 delivery model records a download grant per `(agentId, filename, userId)` in `agent_delivered_files` (`packages/web/src/db/schema.ts`). The serving route (`packages/web/src/app/api/agents/[agentId]/artifacts/[filename]/route.ts`) authorizes by grant lookup, then streams the **current bytes on disk** for that filename. The workspace is shared per agent, not per conversation — on a shared agent, every member's chats write into the same `workbench/` and `uploads/`.

That makes the grant a pointer to a *name*, while the security property we actually promise is about *bytes*: "user U may fetch the file that was delivered to them". Anything that changes the bytes behind a granted name after delivery silently swaps what the grant serves — including content from **another user's** conversation (cross-user exposure on shared agents).

PR #893 recognized this and made `pinchy_generate_file` never-overwrite (`wx` flag + `-1`/`-2` suffix). `email_get_attachment` was already hardened the same way in 3df475b23 (`pickUniqueFilename` + `wx`, TOCTOU-closed, symlink-refusing). So the *tools that mint deliveries* are safe against each other. Two paths remain that can still change bytes under a live grant:

## Gap 1: `pinchy_write` with `overwrite=true`

`packages/plugins/pinchy-files/index.ts` (`pinchy_write`) legitimately supports `overwrite=true` for the agent's working files. But `write_paths` is the same `workbench/` that `pinchy_generate_file` (#893) delivers from. Once #893 merges, a user B conversation can do `pinchy_write(workbench/report.csv, overwrite=true)` and user A's existing `report.csv` download chip now serves B's bytes — or A's regenerated-in-B's-context data, either way not what was delivered to A.

`pinchy_write` itself never emits a `{type:"file"}` artifact block, so it mints no grant — it only mutates bytes behind existing ones.

## Gap 2: cross-zone shadowing (no overwrite flag needed at all)

The grant does not record which zone the file was delivered from, and the serving route searches `DELIVERY_ZONES = ["workbench", "uploads"]` **in order**, serving the first hit:

- User A receives email attachment `invoice.pdf` → bytes in `uploads/invoice.pdf`, grant `(agent, "invoice.pdf", A)`.
- Any later conversation (any member) does a plain **create** `pinchy_write(workbench/invoice.pdf)` — no overwrite flag, no `wx` failure, nothing suspicious.
- A's download now serves `workbench/invoice.pdf` because workbench shadows uploads.

This is the same byte-swap-under-a-grant, reachable without any overwrite semantics.

## Options considered

**(a) Invalidate grants on overwrite.** The overwriting code runs in the plugins (OpenClaw container) which have no DB access; they'd need a new internal API endpoint (Pattern B) called on every overwrite, and a crash between write and invalidate leaves the stale grant serving new bytes. Also poor UX: user A's chip silently starts 404ing. Fail-open design, distributed across processes — not recommended.

**(b) Pin the grant to the delivered bytes: record `zone` + `contentHash` (sha256) at grant time, verify on serve.** *Recommended.*
- `deliverRunArtifacts` (`packages/web/src/server/client-router.ts`) resolves the artifact's on-disk path anyway conceptually; have it locate the file (workbench then uploads, same order as serving), hash it, and store `zone` + `contentHash` on the grant row.
- The serving route serves **only** the recorded zone (kills Gap 2 structurally) and compares the streamed file's hash against the grant (kills Gap 1 and any future write path, including ones we haven't written yet). Mismatch ⇒ 404/410, fail closed.
- Web-process-only change: no plugin changes, no internal API, no distributed invalidation. Overwrites remain allowed for agent working files — they just can't be *served* under an old grant anymore.
- Migration against pre-existing data (per AGENTS.md): existing grants have `zone = NULL, contentHash = NULL`. Decision: **serve them under today's semantics** (search both zones, no hash check) — accept-and-document, with the mandated old-data-new-code test. Alternative (backfill by hashing current disk state) would notarize whatever the bytes happen to be today, which is strictly weaker than honest NULL semantics.
- Known cost: if the *same* user's agent legitimately rewrites a delivered file in place, their old chip goes stale (fail closed). With #893's suffix behavior regeneration produces a new name anyway, so this should be rare; the failure mode is a clear 410, not wrong bytes.

**(c) Suffix behavior for attachment downloads.** Already shipped — `email_get_attachment` has had `pickUniqueFilename` + `wx` since 3df475b23. No action needed.

## Proposed scope

1. Schema: add nullable `zone` + `content_hash` to `agent_delivered_files` + migration.
2. `deliverRunArtifacts`: resolve zone, hash bytes, store both; skip delivery (like the non-servable-MIME case) if the file can't be found/hashed.
3. Serving route: recorded zone only + hash verification when the grant carries one; legacy NULL grants keep current behavior.
4. Tests: route-level (hash match, mismatch ⇒ fail closed, legacy NULL grant still serves, zone pinning), old-data-new-code migration test, E2E download round-trip still green.
5. Docs note in the file-delivery docs.

Should land **after** #893 (same area; #893's `pinchy_generate_file` is what makes workbench grants common).

clemenshelm:
Closes #903.

## The problem

`agent_delivered_files` records a grant per `(agent, filename, user)`, and the serving route streamed **whatever currently sat at that name**. The grant pointed at a *name*; the promise it makes is about *bytes* — "this user may fetch the file that was handed to them". On a shared agent every member's chats write into one workspace, so the two came apart two ways:

| Gap | How | Overwrite flag needed? |
| --- | --- | --- |
| 1 | `pinchy_write(workbench/report.csv, overwrite=true)` in member B's turn replaces the bytes behind member A's live chip | yes |
| 2 | Creating `workbench/invoice.pdf` shadows an `uploads/invoice.pdf` delivery — the route searched zones in order and took the first hit | **no** |

Either way the download served content from a conversation the grantee may have no access to, under their own link, with nothing looking wrong anywhere.

## The fix

A grant now records the **zone** the file came from and the **SHA-256 of its bytes**; the route refuses anything that no longer matches. This is option (b) from the issue, unchanged.

### The hash is the security property — the zone is not

I want to be precise about this rather than claim both halves are load-bearing. Bytes that are not the delivered bytes are refused whatever wrote them and wherever they sit, which covers both gaps and every write path nobody has written yet.

Verified by canary, not by reading:

- deleting the hash comparison → the two tests named after the gaps go **red**
- deleting the zone pin alone → they stay **green**, because a shadowing file with *identical* bytes is harmless

What the zone buys is that a grant's authority is scoped to where its file came from, and that the route looks in one place instead of hashing its way through a search. The comments say exactly that.

### One resolution, not two

`packages/web/src/lib/delivered-file-location.ts` is read by both the side that mints a grant and the side that serves one. A hash taken from one path and checked against another fails open in one direction and 404s every download in the other.

### A grant that cannot be pinned is not minted

`deliverRunArtifacts` locates and hashes before inserting, and skips the delivery otherwise — the same call the non-servable-MIME filter already makes. Minting an unpinnable grant would re-open the gap for *every future* delivery instead of only for the ones that predate this.

## Pre-existing data

Grants written before this carry NULL in both columns and **cannot be pinned honestly** — hashing what is on disk today would certify exactly the swap this prevents. They keep the semantics they were written under.

AGENTS.md § "Test Migrations Against Pre-Existing Data" is discharged by a route-level block that asserts what an upgraded install really does with those rows, including the part nobody wants: *"does not check bytes it was never told about"*. The pin protects deliveries made after the upgrade; the exposure the old ones carry ends when they do.

## Known cost

A delivered file that really is replaced in place now `404`s instead of serving the newer version. With `pinchy_generate_file`'s suffixing (#893) a regeneration produces a new name anyway, so this should be rare — and a clear "not found" beats wrong bytes. Documented in the upgrade note and the file-uploads guide, not hidden.

Serving also costs one extra full read of the file, before a byte goes out. Verifying while streaming would mean the bytes had already left.

## Two things found while writing it

- **`file-uploads.mdx` already promised this.** "Files you were handed earlier — including in other conversations with the same agent — keep exactly the content they had when they were delivered." True of the naming rule it sat beside, false of the product. That sentence is now backed by the code.
- **The open upgrade section's migration list stopped at `0064`** while `0065`–`0069` had landed on main unlisted, so its count was already wrong before `0070` joined. All six are listed now.

## Verification

- `pnpm test` — 11800 passed, 18 skipped, 0 failed
- `pnpm test:scripts` — 0 failed
- `pnpm -C packages/web typecheck`, `pnpm format:check`, eslint on the changed files — clean
- `cd docs && pnpm build && pnpm check:anchors && pnpm check:rendered` — 74 pages, no broken links, every table rendered
- canaries above, run in both directions

`review-docs` was run over the branch.


clemenshelm:
Pushed a follow-up commit (`a6eece9bc`) after reviewing this against #903.

**The pin makes a grant answer for bytes; the delivery glue still asked about a name.** `deliverRunArtifacts` keyed its idempotency set on the filename alone, so a filename that came back carrying *different* bytes read as a repeat poll and minted nothing:

1. the agent generates `report.csv` → grant pinned to those bytes
2. `pinchy_delete` removes it (#1159, merged last night)
3. the agent generates `report.csv` again — `pinchy_generate_file` restarts its collision loop at `attempt === 1`, i.e. at the bare name, so the name comes back
4. `delivered.has(filename)` → skip. No grant, no chip.

and the chip from step 1 now 404s, because its hash names the file step 2 deleted. Before the pin that chip served the new bytes — that was the bug this PR fixes. Refusing them *and* minting nothing is not the fix, it is a dead end with no recovery except a different filename.

Step 2 is why I think this belongs in this PR rather than in a note: deletion only became reachable in #1159. Each change is harmless alone; together they strand the user.

**It also makes this PR's own documentation true.** Both pages tell the reader what to do when a download stops working:

> Ask the agent to hand you the file again and you get a fresh link.

Keyed on the filename, that was precisely the recovery that could not work.

### What changed

The set now maps a filename to the hashes granted for it. A legacy grant carries no hash, so it short-circuits **before** the file is read and keeps the old name-only skip — "did the bytes change?" has no answer for it, and re-polling would append a duplicate row every turn forever.

The cost is honest and worth naming: one re-hash per already-delivered artifact per poll, because the question is now about bytes rather than a name. It is bounded by what the session has delivered, and the serving route already pays the same read on every download.

The serving side needed no change, but it gained the test for the state this makes routine — a stale grant sitting beside a current one for the same filename must serve the file the user currently holds a grant for.

### Verification

The delivery test went red on the current code first (`Number of calls: 0`), then green; the two neighbouring cases — idempotent re-poll and the legacy no-hash grant — stayed green throughout, which is the pair that would break if the new key were simply looser.

Gates on the pushed commit, under Node 22: `pnpm test` 11845 passed / 18 skipped, `pnpm test:scripts` 1240 passed, `test:db` 528 passed across 71 migrations, typecheck and `format:check` clean. One caveat worth stating: the first `test:db` run failed one test against a database it was creating for the first time; the re-run against the settled database is fully green, so I read that as first-run setup rather than a real red.

clemenshelm:
**Note for whoever else is on this branch: I force-pushed over your rebase a few minutes ago. Nothing was lost, but that was luck rather than care, so here is the full account.**

`main` had grown its own `0070` and `0071` (the two `tool_approval` columns), so this branch's `0070_secret_supernaut` collided and the PR went `CONFLICTING` — which is also why no CI ran on my earlier commit: with no merge ref, `pull_request` workflows never start. I rebased onto main, renumbered to `0072`, rebuilt the snapshot from main's `0071` (rather than patching our stale one, which would have silently dropped main's two migrations from the chain), and corrected the migration list in `upgrading.mdx`, which said "Fifteen migrations (`0056`–`0070`)" and named neither of main's.

You did the same thing at the same time and pushed `815ca7c9d`. I then ran `git fetch` immediately before `git push --force-with-lease` — which **defeats the lease**, because the fetch updates the very remote-tracking ref the lease is compared against. So the push went through where it should have been refused.

I diffed the two trees before writing this, rather than assuming they matched. They are functionally identical; the only differences are cosmetic:

| | yours (`815ca7c9d`) | mine (`dc182394c`) |
| --- | --- | --- |
| migration file | `0072_glorious_stryfe.sql` | `0072_secret_supernaut.sql` |
| snapshot `id` | `a8247c8f-…` | `7d60e822-…` (the one this PR was authored with) |
| the `0070`/`0071` docs line | your wording | mine — same facts |

Same `prevId`, same columns, same SQL. Yours came from a fresh `db:generate`; mine keeps the original snapshot id and was verified the same way — `db:generate` reports "No schema changes, nothing to migrate", and `test:db` applies all 73 migrations green (551 tests).

So there is nothing to restore, and re-pushing your version would only churn the branch again. **I am not touching it further.** If you would rather have your revision, force-push it back and I will stay off.

The lesson worth keeping: `--force-with-lease` protects nothing when a `fetch` runs just before it. The lease has to be pinned to what you actually last saw — `--force-with-lease=<ref>:<sha>` with the sha you pushed.

PR #1168, opened 2026-08-06
fix(approvals): wait for OpenClaw's approval id, and deliver the decision when the wait loses

### LINKED ISSUE #1132: Approvals: use OpenClaw's native requireApproval so the run actually pauses
Follow-up to #865 (#124 Tier 2). The gate works and is enforced server-side, but the *interaction* is rough, and all three symptoms share one root cause.

## What's wrong today

We return `block: true` from the `before_tool_call` hook. That correctly prevents the call — but the agent run continues. The model receives our reason as the tool result and talks about it, so the agent answers *before* the user has decided. It reads as if the agent ignored the gate.

Because the run does not pause, approving does nothing on its own: the user has to tell the agent to carry on.

## What OpenClaw already offers

The pinned version (`openclaw@2026.7.1-2`) supports a second return value from the same hook — verified against the docs in the installed bundle (`docs/plugins/hooks.md`, `docs/plugins/plugin-permission-requests.md`):

```ts
requireApproval: {
  title: string;        // ≤ 80 chars
  description: string;  // ≤ 256 chars
  severity?: "info" | "warning" | "critical";
  timeoutMs?: number;   // default 120_000, capped at 600_000
  timeoutBehavior?: "allow" | "deny";
  allowedDecisions?: Array<"allow-once" | "allow-always" | "deny">;
  onResolution?: (decision) => void | Promise<void>;
}
```

> `requireApproval` pauses the agent run and asks the user through plugin approvals.

That is exactly the missing behaviour.

## Why this is reachable now

The 2026-06 assessment concluded this needed `openclaw-node` work. It no longer does:

- `plugin.approval.request | list | waitDecision | resolve` are regular gateway RPCs, and `plugin.approval.requested` / `.resolved` are broadcast events.
- Those events are gated on the `operator.approvals` scope, which is contained in the `operator.admin` scope Pinchy already connects with.
- `openclaw-node` exposes a generic `request(method, params)` and is an `EventEmitter`, so both directions are reachable without a new SDK wrapper.

## Scope

- Hook returns `requireApproval` instead of `block` for gated calls.
- Pinchy listens for `plugin.approval.requested`, renders the card, and resolves via `plugin.approval.resolve`. The run then continues on its own — no "please continue" turn.
- `onResolution` writes the audit row and the consume-once grant. **Pinchy's DB stays the source of truth** for audit, digest binding and durable grants; OpenClaw provides the live mechanics only.
- `allowedDecisions: ["allow-once", "deny"]` — deliberately no `allow-always`, see #NEXT_GRANULAR for the reasoning.

## Known limits, to design around rather than discover

- `timeoutMs` is capped at 10 minutes; our 15-minute TTL has to account for that.
- A paused run does not survive a gateway restart.
- `allow-always` is not durable unless we persist it — which is why we do not offer it.
- If no approval surface is connected, the call is **blocked**, not allowed. Fail-closed, but it means Pinchy must reliably act as that surface.

## Also in scope: the text and the placement

Both change as a consequence, so they belong here rather than in separate PRs.

**Text.** `title` / `description` replace the bare tool name. Sources: `TOOL_REGISTRY` carries `label` and `description`; `summarizeArgs()` already redacts secret-looking keys and truncates long values. Tools outside the registry (`knowledge_search`, the onboarding context tools) need an entry. Research on approval screens is consistent: the reviewer needs the action, the target and the consequence — not the function name.

**Placement.** Inline in the chat thread, at the point the action stands, instead of the fixed card in the bottom-right corner. That is what Claude Code and Cursor do and what the agent-UX guides recommend; a corner card is spatially divorced from the message it belongs to.

clemenshelm:
Closes #1132. Supersedes #1169 — both halves of that race are here, and neither half is sufficient alone.

## The race

OpenClaw mints its approval id **after** `before_tool_call` answers `requireApproval`, and announces it on a separate broadcast. Pinchy's row — and the `approval.requested` audit event the inbox and the E2E both key on — exist **before** that id does. Anything reacting to the card the instant it appears lands inside that gap: the E2E, a scripted approver, a fast human.

Being early was **permanent, not momentary**. `linkApproval` refused a row that was no longer `pending`, and the decision route flips the row before resolving. So the click itself closed the only door the broadcast could still come through. The run then sat until OpenClaw's 600 s cap while the user read *"The agent is no longer waiting for this decision"* — the one thing that was not true.

## Two fixes, because one door is not enough

**1. Wait for the id before settling** (`waitForApprovalLink`, called before the row is flipped). Bounded at 2 s, polled every 50 ms, and it returns immediately whenever nothing can arrive — the id is already there, the row is gone or settled, or it carries no `toolCallId`. So the wait is only ever spent inside the window.

The 2 s is **sized from a measurement**, taken by instrumenting the wait and the bridge and running the integration E2E against the real stack (GitHub Actions was in a major outage and could not answer this):

```
gated call reaches the gate     19:12:13.063
broadcast arrives at Pinchy     19:12:13.282   (+219 ms)
row carries the id              19:12:13.284   (+2 ms, one UPDATE)
waitDecision returns            19:12:13.603   (297 ms parked in total)
```

So the whole hop is ~220 ms locally, and the wait itself measured **0 ms** — the link was already there when the decision arrived. The CI runner that failed was slower: run 31111018560 has the decision POST completing at 14:44:47.827 against the gate's `plugin.approval.request` answered at 14:44:47.944, so the click lost by 25 ms and the link had still not attached ~0.5 s later. 2 s is ~9× the measured chain and ~4× the worst observed shortfall.

**2. Deliver when the wait loses.** `linkApproval` now stores the id on a row that has already been decided, and the bridge hands that decision to the parked call whenever the broadcast arrives — however late.

Without (2), every broadcast past the bound is dropped and the 2 s has a cliff behind it. Without (1), the user is told the decision did not reach the run and only the audit trail corrects it afterwards. Together the fast path is honest and the slow path still completes.

## Gates

`pnpm test` 11 881 passed / 18 skipped · `pnpm -C packages/web test:db` 560 passed (69 files, real PostgreSQL) · `pnpm test:scripts` 1 240 pass / 0 fail · `pnpm build` · `typecheck` + `typecheck:plugins` (10 packages) · `lint` 0 errors · `prettier --check .`

Run on the pinned Node 22. The `pinchy-files` ABI trap AGENTS.md documents bit once on the way here in the other direction — the shared pnpm store's `better-sqlite3` is currently built for NODE_MODULE_VERSION 127, so a Node 24 run reddens 24 tests in `pdf-cache.test.ts` that have nothing to do with this diff.

### The reasoning error was mine, and it was in a test

I wrote the `status = 'pending'` condition and a test asserting it: *"a late broadcast must not re-arm a confirmation the user has already decided"*. That conflates two things:

- `status = 'pending'` is what makes a card **actionable** — it is what `GET /api/approvals` filters on.
- The OpenClaw id is what makes it **resolvable**.

Storing the id on a decided row brings nothing back into the inbox. Refusing to store it threw away the one value the decision needed. The test is **corrected, not deleted**, and now pins both halves: the decided row does take the id, and its status is untouched so the card stays gone. A second test pins the real exception — `consumed`/`expired` are terminal, nothing waits there.

## Observability, which is why this was hard to read

`linkApproval` no longer returns a bare `null` for two very different situations:

- `not-ours` — OpenClaw's own approvals (skill workshop, exec). Silent, as before.
- `settled` — **ours**, and unlinkable. Now says so.

Both were silent before, which is why the first live occurrence took a CI gateway log and a count of RPC lines to diagnose.

And the E2E probe now asserts **the route's own answer first**, as a whole object. The route already knows which of the three branches it took and puts the matching sentence in `resumeError` (nothing-waiting / refused / unreachable); the probe threw that away and asserted the audit row's `outcome`, so a failure read `Expected: "success" / Received: "failure"` and named none of them. Diagnosing one occurrence then cost a gateway log which — it turns out — is only captured on failure at all, so the "control" comparison against a green run was worthless: 6 captured lines against the red run's 387. Silence read as a measurement.

## Coverage where it was missing

Every existing test in "resuming the parked run" called `blockedAndLinked` — it linked **before** deciding. The ordering that broke was the one ordering nothing exercised. There is now an integration test for the reverse order, a route test that lands the broadcast *mid-decision*, and five service tests for `waitForApprovalLink` (four pinning the immediate exits).

**Canary-verified rather than read:** the route test is red without the wait (`expected "vi.fn()" to be called with 'plugin.approval.resolve'`), and stubbing the wait to return `null` reddens two service tests.

The E2E that caught this is **unchanged**, deliberately: its assertion and the product's guarantee are now the same condition, so a test that waited longer than the route does would hide the regression it exists to catch.

## Stated plainly

**Why the id was missing in that particular CI run is not established.** Two candidates: the broadcast had not been processed yet, or it was processed and matched nothing. The gateway log does not record outgoing event frames, so that run cannot decide between them. This PR fixes what is provable by reading — the permanent closure — and makes the next occurrence readable instead of silent.

In the window where the wait loses, the user has already been told the decision did not reach the run, so the message is *pessimistic* rather than wrong-in-their-favour. The run continues, and `onResolution` records what the runtime actually did.

### DIFF
--- a/packages/web/e2e/integration/agent-chat.spec.ts
+++ b/packages/web/e2e/integration/agent-chat.spec.ts
@@ -765,6 +765,21 @@ test.describe("Plugin behavior — pinchy-approvals", () => {
       });
       expect(decision.status()).toBe(200);
 
+      // Assert the ROUTE'S OWN ANSWER first, and assert the whole object.
+      //
+      // The route already knows why a decision did not reach the parked call —
+      // it puts `resumeError` in this body, one string per branch
+      // (nothing-waiting / refused / unreachable). The first version of this
+      // probe threw that away and asserted only the audit row's `outcome`, so
+      // a failure read `Expected: "success" / Received: "failure"` and named
+      // none of the three. Diagnosing one occurrence then took the CI gateway
+      // log — which, it turns out, is only captured on failure at all, so the
+      // "control" comparison against a green run was worthless.
+      //
+      // toMatchObject on the whole body is what surfaces it: an unexpected
+      // `resumeError` shows up in the diff instead of being invisible.
+      expect(await decision.json()).toMatchObject({ ok: true, resumed: true });
+
       // The decision is recorded as approval.granted — and `resumed` says it
       // reached the parked call. Asserting only that the row exists would pass
       // just as happily over a run that stays parked until it times out, which
@@ -774,8 +789,12 @@ test.describe("Plugin behavior — pinchy-approvals", () => {
         predicate: (e) => e.resource === `approval:${requestId}`,
         deadlineMs: 15000,
       });
-      expect(granted.outcome).toBe("success");
-      expect((granted.detail as { resumed?: boolean }).resumed).toBe(true);
+      // Same reasoning: compare the fields together so a failure prints the
+      // recorded reason rather than just the word "failure".
+      expect({
+        outcome: granted.outcome,
+        ...(granted.detail as { resumed?: boolean; resumeReason?: string; resumeDetail?: string }),
+      }).toMatchObject({ outcome: "success", resumed: true });
 
       // The runtime reports back that it let the call through…
       await pollAuditForEvent(page, {
--- a/packages/web/src/__tests__/api/approvals.integration.test.ts
+++ b/packages/web/src/__tests__/api/approvals.integration.test.ts
@@ -462,33 +462,71 @@ describe("approval routes (integration, real DB)", () => {
       expect(entry.detail).toMatchObject({ resumed: false });
     });
 
+    // The id arrives on a broadcast OpenClaw only sends AFTER the gate answered
+    // `requireApproval`, so a decision taken the moment the card appears lands
+    // before it. That used to report `nothing-waiting` — a false statement
+    // about a call that IS parked — and it was permanent, because linkApproval
+    // refuses a row that is no longer pending: the run then sat until
+    // OpenClaw's 600 s cap, whatever the user clicked. CI, 2026-08-06.
+    it("resumes a decision taken before the approval broadcast arrived", async () => {
+      const blocked = await (
+        await gateCheck(gateReq({ ...gateBody(), toolCallId: "call_race" }))
+      ).json();
+      setSession(user);
+
+      // The broadcast lands while the decision is already in flight.
+      const linked = new Promise<void>((resolve) => {
+        setTimeout(() => {
+          void linkApproval({ approvalId: "plugin:late", toolCallId: "call_race" }).then(() =>
+            resolve()
+          );
+        }, 150);
+      });
+
+      const res = await decide(decideReq({ decision: "approve" }), ctx(blocked.requestId));
+      await linked;
+
+      expect(gatewayRequest).toHaveBeenCalledWith("plugin.approval.resolve", {
+        id: "plugin:late",
+        decision: "allow-once",
+      });
+      expect(await res.json()).toMatchObject({ ok: true, resumed: true });
+
+      await flushAfter();
+      const [entry] = await db
+        .select()
+        .from(auditLog)
+        .where(eq(auditLog.eventType, "approval.granted"));
+      expect(entry.outcome).toBe("success");
+      expect(entry.detail).toMatchObject({ resumed: true });
+    });
+
     /**
-     * The ordering every other test in this block assumes away.
+     * And when the broadcast misses even the wait.
      *
-     * `blockedAndLinked` links BEFORE deciding, which is the common case and
-     * was the only one covered. `gate-check` creates the row before the hook
-     * returns `requireApproval`, though, so the card is clickable before
-     * OpenClaw has announced the approval — and a decision taken in that window
-     * had no id to resolve with. It used to stay that way permanently, because
-     * `linkApproval` only matched `pending` rows and `resolveDecision` had
-     * already moved the row out of `pending`. The run then sat parked until
-     * OpenClaw's 600 s cap.
+     * The wait above is bounded, so it can still lose — and losing it used to
+     * be PERMANENT: `linkApproval` matched `pending` only, while the route had
+     * already flipped the row, so the id could never attach afterwards and the
+     * run sat until OpenClaw's 600 s cap with nothing saying why.
      *
-     * Caught in CI on 2026-08-06 by the E2E probe, and only by luck of timing:
-     * one plugin.approval.request, zero plugin.approval.resolve.
+     * This asserts the contract the bridge stands on rather than the resume
+     * itself: the link still attaches to a decided row, and it carries the
+     * decision in `status`. What the bridge then does with that is pinned in
+     * `plugin-approval-bridge.test.ts` ("delivers a decision that was made
+     * before the approval was announced") — deliberately not restated here,
+     * because a test that names a guarantee it does not exercise is worse than
+     * no test at all.
      */
-    it("resumes a run whose approval was announced after the user had already decided", async () => {
+    it("leaves a decided row linkable, so a broadcast past the wait still has somewhere to land", async () => {
       const blocked = await (
         await gateCheck(gateReq({ ...gateBody(), toolCallId: "call_early" }))
       ).json();
       setSession(user);
 
-      // Decide first — no id on the row yet.
+      // Decide first — no id on the row, and none arrives inside the wait.
       await decide(decideReq({ decision: "approve" }), ctx(blocked.requestId));
       expect(gatewayRequest).not.toHaveBeenCalled();
 
-      // …then the broadcast lands. This is the only thing left that can carry
-      // the decision to the parked call.
       const outcome = await linkApproval({
         approvalId: "plugin:late",
         toolCallId: "call_early",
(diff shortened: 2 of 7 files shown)


clemenshelm:
## Correction: this PR does not fix the CI failure, and the diagnosis that motivated it was partly unfounded

Recording this against the PR rather than quietly force-pushing over it.

**What I claimed:** the integration probe fails because the user's decision arrives before OpenClaw's `plugin.approval.requested` broadcast has been linked, so `resolvePluginApproval` short-circuits on a null id.

**What actually supports that:** in the failing run's gateway log, `plugin.approval.request` appears once and `plugin.approval.resolve` never, with no `[approvals]` line — inside a capture window that covers the failure. That is suggestive, not conclusive.

**What does not support it:** I used the green run on #1158 as a control ("the probe passed there, so this is intermittent, not broken"). Before saying that I should have checked whether the green run logged anything at all. It did not:

| run | captured `openclaw-1` lines |
| --- | --- |
| green (#1158, job 92445802641) | 6 |
| red (#1168, job 92649882069) | 387 |

The gateway log is only dumped on failure. So the absence of `plugin.approval.resolve` in the green run is not a measurement, and the control was worthless.

**What this PR still fixes, independently of that incident** — proven by reading the code and verified against real PostgreSQL, not inferred from CI:

`linkApproval` matched `WHERE toolCallId = ? AND status = 'pending'`, while `resolveDecision` flips the row to `approved`/`denied` before the route reads the id. So the click closed the window itself: a broadcast arriving even a millisecond later could never land, the row would never be `pending` again, and the run stayed parked until OpenClaw's 600 s cap with nothing recording why. It also removes a genuine ambiguity — "this approval is not ours" and "this approval is ours but unlinkable" both returned a bare `null` and both were silent.

I also had that defect pinned as a test asserting the wrong invariant, with a rationale I wrote ("a late broadcast would make a settled confirmation look resolvable again"). It is wrong: `status` is what `GET /api/approvals` filters on and therefore what makes a card actionable; the OpenClaw id is what makes it *resolvable*. That test is corrected, not deleted, and the correction is argued in the commit.

**Why it does not turn the probe green even if the race is real:** the probe asserts the audit row written at decision time. Event-driven late delivery resumes the run but leaves that row saying `resumed: false`. I defended that tradeoff as "pessimistic message, but the run continues" — too weak for an audit trail, which would then be permanently wrong about a case that worked.

**What I am doing instead of a third fix aimed at a fourth hypothesis:** the latest commit makes the probe assert the route's own answer, which already carries `resumeError` naming exactly which of `nothing-waiting` / `refused` / `unreachable` occurred. The next failure states its cause instead of requiring log archaeology. The bounded wait in the decision route comes after that answer, not before it.

clemenshelm:
## Superseded by #1167 — recommend closing

#1167 landed the same defect class on main, and went further than this branch did:

- `linkApproval` now accepts `["pending", "approved", "denied"]` — the core fix here, that a decided row must still be able to receive OpenClaw's id.
- Plus `isNull(openclawApprovalId)` and session scoping, which this branch does not have.
- Plus `awaitApprovalLink` with `LINK_WAIT_MS = 2000`, **sized from a measurement** ("one `plugin.approval.request` round trip, ~125 ms on a loaded CI runner") rather than from my guess.
- Plus `e601d693b`, letting a retried call's approval id replace a dead one — a case I had not considered at all.

Rebasing this branch onto main would therefore resolve four conflicts by taking main's side in nearly all of them, and leave a PR whose remaining content is one test file.

**That one file is genuinely still missing on main**, so I lifted it onto current main as #1171 rather than rebasing this: the probe improvement that makes the failure name its own cause. Main's copy of `agent-chat.spec.ts` still asserts only `granted.outcome`, which is what made the original failure opaque.

Leaving the close to you, since you pushed the last three commits here.

For the record, the correction from [my earlier comment](https://github.com/heypinchy/pinchy/pull/1168#issuecomment-5208204434) stands and is worth keeping visible: I used the green #1158 run as a control without checking whether it had logged anything, and it had not — the gateway log is only dumped on failure. Two fixes got aimed at hypotheses that evidence never supported. #1171 is the part that makes the next one self-explaining.

clemenshelm:
Superseded by `6b58b9113` on main, which landed with #1167 while this was waiting on CI.

That commit is the same fix, in both halves — `linkApproval` records the id on an already-decided row, and the decision route waits, bounded, for the id when it is not there yet (`awaitApprovalLink`, 2 s at 40 ms; this branch had 2 s at 50 ms). It reaches the same diagnosis this PR did, down to the reason the miss was permanent rather than momentary.

Rebasing was not worth it: against current main this branch is **+620 / −979**, and it would carry #1167's session-scope fix and #1163's KB-abstention fix back out with it. The red required check said so in its own words — the test-removal guard read the 15 tests main had *gained* as removals, and its warning names exactly that case.

One thing main deliberately does not do, recorded here rather than lost: the bridge discards its link result (`void linked`), so a broadcast arriving after the 2 s bound leaves the decision undelivered. Main's own commit message is explicit that it "ends in an honest no" there. This branch additionally called `resolvePluginApproval` from the bridge in that case. Measured, the window is narrow — the hop is ~220 ms locally and ~0.5 s on the CI runner that failed, against a 2 s bound — so the cliff is being accepted knowingly, not overlooked.

Closing unmerged. #1169 (the wait half alone, without the late delivery) is already closed.

PR #1169, opened 2026-08-06
fix(approvals): wait for OpenClaw's approval id before settling the decision

clemenshelm:
## What went red

`Integration Tests (OpenClaw + Fake Ollama) (1/2)` failed on #1164 — a PR that changes four comments and a drift guard, so not its doing:

```
✘ Plugin behavior — pinchy-approvals › a gated tool is blocked into an approval request the user can grant
  expect(granted.outcome).toBe("success")
  Expected: "success"   Received: "failure"
```

Read from the failure trace rather than guessed:

```json
{ "resumed": false, "resumeReason": "nothing-waiting" }
```

`nothing-waiting` about a call that was demonstrably parked — the gateway answered `plugin.approval.request` 0.6 s earlier, in the same log.

## Why it happens

The window is **structural**, not a hiccup. OpenClaw mints its approval id only *after* `before_tool_call` answers `requireApproval`, and announces it on a separate broadcast. Pinchy's own row — and the `approval.requested` audit event the inbox and the E2E both key on — exist **before** that id does. Anything that reacts to the card the instant it appears lands inside the gap: the E2E, a scripted approver, a fast human.

And being early was **permanent, not momentary**. `linkApproval` refuses a row that is no longer `pending`, and the route flips the row before resolving. So a link in flight when the decision arrived could never attach afterwards: the run sat until OpenClaw's 600 s cap, while the user read *"The agent is no longer waiting for this decision"* — the one thing that was not true.

## The fix

`waitForApprovalLink`, called **before** the row is flipped so the link can still land. Bounded at 2 s, polled every 50 ms.

It returns immediately whenever nothing can arrive — the id is already there, the row is gone or settled, or it carries no `toolCallId` (the key a broadcast is matched on). So the wait is only ever spent inside the race window. It is scoped to the caller's own row, so a stranger's request does not become measurably slower than an unknown one; `resolveDecision` still owns the 403.

### About the 2 s — measured, not guessed

An earlier draft of this PR overclaimed this, so here is the actual measurement. GitHub Actions is in a [major outage](https://www.githubstatus.com/) today and could not answer it, so I instrumented `waitForApprovalLink` and the bridge and ran the integration E2E against the real stack (OpenClaw container, fake Ollama) locally:

| time | event |
|---|---|
| `19:12:13.063` | gated call reaches the gate |
| `19:12:13.282` | broadcast arrives at Pinchy (**+219 ms**) |
| `19:12:13.284` | row carries the id (**+2 ms**, one UPDATE) |
| `19:12:13.603` | `waitDecision` returns — **297 ms** parked in total |

The wait itself measured **0 ms**: locally the link is already there when the decision arrives. The CI runner that failed was slower — there the link had still not attached ~0.5 s after the request RPC was answered.

So 2 s is **~9x the measured chain** and **~4x the worst observed shortfall**, while staying short enough that no decision feels hung. Instrumentation was removed before pushing. If the bound ever does prove short, the symptom is the same honest `resumed: false` it replaced, from one place.

The E2E itself **passes against the real stack** (9.7 s) — the end-to-end proof CI owes us once it is back.

## Tests

The E2E that caught this is **unchanged**, and that is deliberate rather than lazy: its assertion and the product's guarantee are now the same condition, so a test that waited longer than the route does would hide the regression it exists to catch.

Added:

- a route test that lands the broadcast *mid-decision* and asserts the run resumes with `outcome=success` / `resumed: true`;
- five service tests for `waitForApprovalLink`, four of them pinning the immediate exits (already linked, no `toolCallId`, settled, not the caller's).

**Canary-verified rather than read:** the route test is red without the wait (`expected "vi.fn()" to be called with 'plugin.approval.resolve'`), and stubbing the wait to return `null` reddens two service tests.

## Stated plainly

A broadcast arriving **after** the 2 s bound is still dropped — nothing re-resolves a settled row. Closing that would mean resuming out of band and writing a second audit row, which is a bigger change than this one.

## Gates

`pnpm test` (11 877 passed / 18 skipped) · `pnpm -C packages/web test:db` (558 passed, 69 files) · `pnpm test:scripts` (1 240) · `pnpm -C packages/web typecheck` · `pnpm -C packages/web lint` (0 errors) · `prettier --check .`

### DIFF
--- a/packages/web/src/__tests__/api/approvals.integration.test.ts
+++ b/packages/web/src/__tests__/api/approvals.integration.test.ts
@@ -462,6 +462,45 @@ describe("approval routes (integration, real DB)", () => {
       expect(entry.detail).toMatchObject({ resumed: false });
     });
 
+    // The id arrives on a broadcast OpenClaw only sends AFTER the gate answered
+    // `requireApproval`, so a decision taken the moment the card appears lands
+    // before it. That used to report `nothing-waiting` — a false statement
+    // about a call that IS parked — and it was permanent, because linkApproval
+    // refuses a row that is no longer pending: the run then sat until
+    // OpenClaw's 600 s cap, whatever the user clicked. CI, 2026-08-06.
+    it("resumes a decision taken before the approval broadcast arrived", async () => {
+      const blocked = await (
+        await gateCheck(gateReq({ ...gateBody(), toolCallId: "call_race" }))
+      ).json();
+      setSession(user);
+
+      // The broadcast lands while the decision is already in flight.
+      const linked = new Promise<void>((resolve) => {
+        setTimeout(() => {
+          void linkApproval({ approvalId: "plugin:late", toolCallId: "call_race" }).then(() =>
+            resolve()
+          );
+        }, 150);
+      });
+
+      const res = await decide(decideReq({ decision: "approve" }), ctx(blocked.requestId));
+      await linked;
+
+      expect(gatewayRequest).toHaveBeenCalledWith("plugin.approval.resolve", {
+        id: "plugin:late",
+        decision: "allow-once",
+      });
+      expect(await res.json()).toMatchObject({ ok: true, resumed: true });
+
+      await flushAfter();
+      const [entry] = await db
+        .select()
+        .from(auditLog)
+        .where(eq(auditLog.eventType, "approval.granted"));
+      expect(entry.outcome).toBe("success");
+      expect(entry.detail).toMatchObject({ resumed: true });
+    });
+
     it("does not call the gateway when no run is parked on that call", async () => {
       // No approval id on the row: the gate refused without suspending, the
       // approval already timed out, or the row predates #1132.
--- a/packages/web/src/app/api/approvals/[id]/decision/route.ts
+++ b/packages/web/src/app/api/approvals/[id]/decision/route.ts
@@ -3,7 +3,7 @@ import { eq } from "drizzle-orm";
 import { withAuth } from "@/lib/api-auth";
 import { parseRequestBody } from "@/lib/api-validation";
 import { decisionSchema } from "@/lib/schemas/approvals";
-import { resolveDecision } from "@/lib/approvals/service";
+import { resolveDecision, waitForApprovalLink } from "@/lib/approvals/service";
 import { resolvePluginApproval } from "@/server/resolve-plugin-approval";
 import { type AuditLogEntry } from "@/lib/audit";
 import { deferAuditLog } from "@/lib/audit-deferred";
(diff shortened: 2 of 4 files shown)


clemenshelm:
Verified locally while GitHub Actions is in a major outage — the full integration suite (both CI shards' worth) against the production images:

```
40 passed (7.5m)
✓ 27 agent-chat.spec.ts:711 › Plugin behavior — pinchy-approvals ›
     a gated tool is blocked into an approval request the user can grant (6.8s)
```

That is the exact test that went red on #1164, plus every other integration spec, green on this branch. Stack: `docker-compose.yml + e2e + integration`, real OpenClaw container, `Dockerfile.pinchy` image.

Not a substitute for CI, and I am not claiming it is — CI still has to run once GitHub recovers. But it is the same suite on the same images, so the fix is not resting on unit tests alone.

clemenshelm:
Re-triggering CI: the run for `989cf4837` was never created — its webhook was lost in yesterday's GitHub Actions outage, and the only run on this branch (for the previous commit) had every job cancelled after 3+ hours in the queue without executing. Reopening immediately.

clemenshelm:
Superseded by `6b58b9113` on main (plus `530219203`, `feb8c15bb`, `e601d693b`) — closing.

Same defect, same diagnosis, arrived at independently: `approval.requested` is written from inside gate-check, so the card is actionable a full `plugin.approval.request` round trip before Pinchy learns the id, and a decision in that window reported `nothing-waiting` about a run that was still parked. Both fixes land a bounded wait at 2 s (main polls at 40 ms, this branch at 50 ms).

Main's version is a strict superset, and I would rather have it than mine:

- It closes the tail this PR explicitly left open. `linkApproval` now records the id on an already-decided row, so a broadcast that arrives after a fast decision still lands and the decision can be delivered. My PR named that as "not fixed here".
- It waits **after** `resolveDecision` rather than before. That is the better ordering: the 403 is answered before any waiting, so there is no timing side-channel — I had to scope my wait to `requesterId` to mitigate a problem main does not have.
- It also fixes two things this branch never touched: session-scoped linking (self-hosted OpenAI-compatible servers number tool calls per response, so `call_0` can collide across users) and re-pointing a retried call's approval id.

The measurements agree, which is a small comfort: main measured the gap at ~125 ms on the failing CI run; instrumenting the real stack locally here gave ~220 ms broadcast + 2 ms UPDATE, 297 ms parked in total. Same order, same conclusion about a 2 s bound.

One observation from this branch that main does not cover, offered rather than pushed: `awaitApprovalLink` burns the full 2 s on a row that carries no `toolCallId`, which no broadcast can ever name (the "still decides a gated call that carries no toolCallId" path). An early exit there would save a pointless wait on those decisions. Happy to file it if it is worth having.

PR #1172, opened 2026-08-07
fix(deps): pdfjs-dist 6.2.108 closes the PDF XSS advisory

clemenshelm:
The third of the advisories `osv-scan` reports, and the only one that is not a lockfile-only move. The other two (mermaid, js-yaml) are in #1170.

## The advisory

**GHSA-hq66-cqwq-w95j / CVE-2026-16633, CVSS 8.6.** A crafted PDF can run attacker-controlled JavaScript in the hosting domain's context when scripting is enabled and no CSP intervenes. Affected range is 5.6.83 up to 6.2.108.

**There is no 5.x fix.** 5.7.284 — the version pinned here — is the last 5.x ever published, so the major bump is the only way out. Both declaring packages move together: `pinchy-files` (extraction) and `packages/web` (knowledge-base ingestion).

## How exposed is Pinchy, really

Stated plainly, because it governs how the 8.6 should be read: **pdfjs runs server-side only here.** `getDocument` from the legacy build, for text and image extraction; `next.config.ts` keeps it out of the client bundle via `serverExternalPackages`. No browser viewer renders PDFs through pdf.js, so the advisory's stated impact — XSS in the hosting domain — does not describe this deployment.

The bump is still the right call. That mitigation rests on a property of our code that **no test asserts**, and a future in-browser viewer would silently inherit the hole.

## The breaking change we hit

pdfjs 6 removes `PDFDocumentProxy.destroy()`. Four call sites used it, two of them production code:

| File | |
| --- | --- |
| `packages/plugins/pinchy-files/pdf-extract.ts` | production |
| `packages/web/src/lib/knowledge/pdf-extract.ts` | production |
| `packages/plugins/pinchy-files/pdf-render.test.ts` | test |
| `packages/plugins/pinchy-files/generate-file.test.ts` | test |

Each now keeps the loading task and tears down through it — which is what the removed method delegated to anyway. **No error path changes shape**: web keeps its `try/finally`, pinchy-files keeps its straight-line teardown. I did not "improve" either while passing through.

Verified empirically rather than from the release notes: the legacy build still exists in 6.2.108, and `loadingTask.destroy` is a function while `PDFDocumentProxy.destroy` is gone — the typecheck named all four sites before the change and none after.

## A duplicate drops out

The lockfile loses 139 lines. pdfjs 5 pulled `@napi-rs/canvas` **0.1.100** in transitively, a second copy beside the **1.0.3** that `pinchy-files` declares directly. pdfjs 6 does not, so the duplicate (and its eleven platform binaries) is gone. Rendering still runs through the plugin's own `CanvasFactory` — `pdf-render.test.ts` exercises that for real, against a fixture PDF.

## Verification

- `pnpm test` — **11893 passed** across 822 files
- `pnpm -C packages/plugins/pinchy-files test` — 259 passed
- `pnpm typecheck:plugins` — 0 errors
- `pnpm -C packages/web typecheck` — 0 errors
- `pnpm format:check`

all exit 0.

### DIFF
--- a/packages/plugins/pinchy-files/generate-file.test.ts
+++ b/packages/plugins/pinchy-files/generate-file.test.ts
@@ -7,15 +7,16 @@ import { generateFile } from "./generate-file";
  * by an unbounded wrap (the regression #1/#2 guard against). No font data or
  * canvas factory is needed just to read structural page count. */
 async function countPdfPages(buffer: Buffer): Promise<number> {
-  const doc = await getDocument({
+  const loadingTask = getDocument({
     data: new Uint8Array(buffer),
     isEvalSupported: false,
     disableAutoFetch: true,
     disableFontFace: true,
     useSystemFonts: false,
-  } as Record<string, unknown>).promise;
+  } as Record<string, unknown>);
+  const doc = await loadingTask.promise;
   const numPages = doc.numPages;
-  await doc.destroy();
+  await loadingTask.destroy();
   return numPages;
 }
 
--- a/packages/plugins/pinchy-files/package.json
+++ b/packages/plugins/pinchy-files/package.json
@@ -12,7 +12,7 @@
     "better-sqlite3": "^12.11.1",
     "exceljs": "^4.4.0",
     "mammoth": "^1.12.0",
-    "pdfjs-dist": "^5.7.284",
+    "pdfjs-dist": "^6.2.108",
     "pdfkit": "^0.19.1",
     "turndown": "^7.2.0",
     "turndown-plugin-gfm": "^1.0.2"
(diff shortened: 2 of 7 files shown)


clemenshelm:
Superseded by #1170, which now carries this exact commit (`fix(deps): pdfjs-dist 6.2.108 closes the PDF XSS advisory`) unchanged, cherry-picked on top of the mermaid and js-yaml fixes.

Reason for folding them together: `osv-scan` reports the whole tree, not the diff. Two independent branches off `main` each saw the other's findings, so both stayed red however correct each one was — this PR reported mermaid and js-yaml, #1170 reported pdfjs. Neither was wrong; a repo-wide check simply cannot go green while its fixes live on separate branches.

Nothing is lost: same commit, same verification, and the combined branch was re-tested from scratch after the rebase (11893 tests, both lockfiles clean under `--frozen-lockfile`).

